PR #8, opened 2020-04-19
Provide a yarn add step for gem plugins with the relevant metadata when they get loaded

jaredcwhite:
This lets you use an NPM package which will get installed and run alongside your plugin. The way it works is, from the docs:

----

Let's assume you've been building an awesome plugin called, unsurprisingly,
`MyAwesomePlugin`. In your `my-awesome-plugin.gemspec` file, all you need to do is
add the `yarn-add` metadata matching the NPM package name and keeping the version
the same as the Gem version:

```ruby
  spec.metadata = { "yarn-add" => "my-awesome-plugin@#{MyAwesomePlugin::VERSION}" }
```

With that bit of metadata, Bridgetown will know always to look for that package in
the users' `package.json` file when they load Bridgetown, and it will trigger a
`yarn add` command if the package and exact version number isn't present.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/plugin_manager.rb
+++ b/bridgetown-core/lib/bridgetown-core/plugin_manager.rb
@@ -26,6 +26,7 @@ module Bridgetown
 
         Bundler.setup
         required_gems = Bundler.require(:bridgetown_plugins)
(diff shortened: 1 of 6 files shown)


jaredcwhite:
Ship it! :shipit:

PR #14, opened 2020-04-22
Update sass-loader webpack config to support .sass

jaredmoody:
I may be _completely_ crazy, but I like the indented form of sass and the `.sass` file extension.

This little tweak (ripped from the [sass-loader readme](https://github.com/webpack-contrib/sass-loader)) makes it so webpack supports both out of the box.


jaredcwhite:
Totally, that should work!

jaredcwhite:
Wow, I am a dummy! I though I'd merged this in, but actually I'd only closed the PR. Let's try this again!!!

PR #21, opened 2020-04-25
Add Concurrently and Browsersync for live reload

jaredcwhite:
This PR fits in the final piece of the Webpack puzzle for brand-new site builds by providing a single command/script: `yarn start`

This kicks off both the Webpack and Bridgetown watchers (plus WEBrick running on port 4001) and then fires up Browsersync to service port 4000 as a live-reload dev server. Everything's managed by Concurrently in a single terminal instance. It's truly a thing of beauty! 😆


jaredcwhite:
LGTM 👍

PR #26, opened 2020-04-28
Proof of concept "Liquidbook" — first implementation of Liquid Components 

jaredcwhite:
In a first "proof of concept" pass at implementing the [Liquid Components spec](https://github.com/bridgetownrb/liquid-component/blob/master/README.md), this PR allows YAML front matter to live at the top of Liquid templates for use with the `render` tag. For example:

```liquid
---
name: Back to Button
description: Display a button in the upper-right corner of the content reigion to go back to a previous section.
variables:
  url: ["url", "The address of the page to go to"]
  label: ["string", "The label of the button"]
---
<p class="has-text-right is-pulled-right mt-1 pl-5 pb-5">
  <a href="{{ url }}" class="button is-info is-small">
    <span class="icon" aria-hidden="true">
      <i class="fa fa-chevron-up"></i>
    </span>
    <span>{{ label }}</span>
  </a>
</p>
<div class="is-clearfix is-hidden-tablet"></div>
```

Our implementation of the Liquid File System which finds and loads in templates will automatically detect and strip out YAML front matter and provide just the actual template code to the Liquid parser.

Doing something extra with that front matter isn't currently a feature of Bridgetown Core. However, the public website now has a custom plugin and set of templates and other logic to build up a "Liquidbook" UI (aka Storybook for Liquid Components) to list out and preview components. It's available via this PR at `/components`. At some point in the future I imagine it'll make sense to extract a bunch of this out into its own project/gem for others to use, but this is just the initial proof of concept.

**Check it out in the deploy preview!**
https://deploy-preview-26--bridgetownrb.netlify.app/components/


bridgetown-core/lib/bridgetown-core/liquid_renderer/file_system.rb
```diff
@@ -24,13 +37,35 @@ def read_template_file(template_path)
...
+        file_content = ""
 
-        # Last path in the list wins
-        ::File.read(found_paths.last)
+        # Strip YAML header
+        if Utils.has_yaml_header?(filename)
+          begin
+            content = ::File.read(filename)
+            file_content = $POSTMATCH if content =~ Document::YAML_FRONT_MATTER_REGEXP
+          rescue Psych::SyntaxError => e
```
jaredcwhite, line 52: I just realized we don't need this since we're not actually parsing the YAML, just stripping it out :)

jaredcwhite:
TODO: I'm now thinking virtually all of this needs to be in separate gem(s). Bridgetown can reference (some of) them as default dependencies of course, but it's really not Bridgetown's responsibility to deal with the low-level machinations of how Liquid Components work.

jaredcwhite:
OK, this is looking pretty good now. It's using the new [Liquid Component](/bridgetownrb/liquid-component) gem to parse components, and the preview logic got refactored into a concise builder plugin. At some point down the road it'll be cool to extract the component previews into its own gem-based plugin, but for this initial pass I think keeping it in the website repo as a demo will be fine.

jaredcwhite:
Haha, it might help if I actually pushed up the commit 🤣

jaredcwhite:
All right, `{% include %}` is officially deprecated now and will be removed in Bridgetown 1.0. All hail our new overlords, the `{% render %}` and `{% rendercontent %}` tags. (Finally!!)

jaredcwhite:
Merged into the 0.15 beta release branch.

PR #30, opened 2020-05-01
Add Faraday to default gems

### LINKED ISSUE #19: feat: Switch tests from using httpclient to Faraday (and then future data importers can expect Faraday out-of-the-box)
## Summary

I'm currently reviewing Bridgetown's plugins/hooks system and iterating towards an out-of-the-box solution for external API data import starting with REST-style calls. That let me to review what, if anything, we're currently using for an HTTP client. Currently there are some `serve` commands that seem to use the `httpclient` gem, so that's a good place to start refactoring.

## Motivation

My ultimate goal is to allow folks to write tiny plugins (aka barely more than a one-liner in many cases) that will automatically provide data from external APIs to the site build (pages, templates, etc., so it's important that Bridgetown come with an opinionated default of an HTTP client plugins can standardize on. And for a variety of reasons (not least of which it's part of the standard Rails bundle), Faraday seems top of the list. The API for instantiating connections can get rather verbose, but we can easily built a thin wrapper around that which will be sufficient for "most" use cases.

jaredcwhite:
Add Faraday to the default set of gems that get installed with Bridgetown. Now site builders and gem plugins can all expect a modern Ruby HTTP client with all the bells and whistles ready to go out of the box.

Closes #19

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,6 @@ end
 
 group :test do
   gem "cucumber", "~> 3.0"
-  gem "httpclient"
   gem "jekyll_test_plugin"
   gem "jekyll_test_plugin_malicious"
   gem "memory_profiler"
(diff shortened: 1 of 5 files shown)


jaredcwhite:
LGTM

PR #33, opened 2020-05-03
Honor the configured site encoding when loading Liquid components

jaredcwhite:
I ran across a `US-ASCII` encoding bug when using a remote SSH session that was incorrectly configured (aka not indicating a `UTF-8` shell) in conjunction with Bridgetown's new support for Liquid's `render` tag. It was preventing the site from building.

This resolves that error.


jaredcwhite:
LGTM :shipit:

Issue #31, opened 2020-05-02
Include ActiveSupport and refactor accordingly


jaredcwhite:
I've come to discover I'm instinctually an "ActiveSupport-flavored" Rubyist, so not having all the niceties that come from ActiveSupport is a real bummer. I've already identified a number of places where both the Bridgetown internals and external plugin APIs could be improved substantially from utilizing ActiveSupport, so I think it's time to bite the bullet and get this added. I'm not sure if it will add a small amount of extra boot time, so that's something to consider, but if we're talking 100ms or less, I'm not too worried.

(Another good reason to add ActiveSupport: a major goal for this project is to find symbiotic ways for Bridgetown and Rails to work together and swim in the same developer communities, so the less friction there is for Rails folks to get up to speed with Bridgetown, the better.)

jaredcwhite:
Note: changing milestone to 0.14 due to using HashWithIndifferentAccess for upcoming plugin DSL stuff, but really the bulk of the work to implement this will be in 0.15 and beyond.

PR #38, opened 2020-05-08
Add titleize Liquid filter

jaredcwhite:
Like the docs say: a Liquid filter to transform a lowercase string, slug, or identifier string into a capitalized title. Uses the already present `Utils.titleize_slug` method.

input: `{{ "to-kill-a-mockingbird" | titleize }}`
output: `To Kill A Mockingbird`

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/filters.rb
+++ b/bridgetown-core/lib/bridgetown-core/filters.rb
@@ -63,6 +63,16 @@ module Bridgetown
       Utils.slugify(input, mode: mode)
     end
 
+    # Titleize a slug or identifier string.
+    #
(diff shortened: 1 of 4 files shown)


westonganger:
I dont like that this method converts dashes to spaces. I think that is undesirable for a general titleize method.

jaredcwhite:
@westonganger Well that's how `titleize` works currently in ActiveSupport which we're hoping to integrate more of per #31, and it's also how Jekyll has worked in the past. Were you looking for something that only coverts underscores to spaces?

westonganger:
After considering further hyphenated words are rarely necessary. Might as well stick to the convention.

PR #39, opened 2020-05-09
add a posts page to the new site template

andrewmcodes:

- [ ] I've added tests (if it's a bug, feature or enhancement)
        __I did some looking and I did not find where you were testing this. It is totally possible, if not probable, that I just missed it so would appreciate you pointing me in that direction if I did indeed just overlook it :smile:__
       __I did some looking and I do not see it documented what get's generated from the build command__

## Summary

When generating a new bridgetown site, it is not immediately clear how to get to your posts. I attempted a few ways to navigate there, but since I do not have prior experience with Jekyll, I was a little lost. This commit adds a posts page so that feature newbies, like myself, can quickly see their posts, and not have to try and spend time figuring it out.

## Context

Not related to any currently open issues, but may help prevent some from coming down in the future from people who get immediately frustrated and don't take the time to investigate.

### DIFF
--- a/bridgetown-core/lib/site_template/src/_includes/navbar.html
+++ b/bridgetown-core/lib/site_template/src/_includes/navbar.html
@@ -1,4 +1,5 @@
 <nav>
   <a href="/">Home</a>
   <a href="/about">About</a>
+  <a href="/posts">Posts</a>
 </nav>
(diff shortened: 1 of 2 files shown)


jaredcwhite:
@andrewmcodes Looks great!

The `new` command is tested on a basic level in `bridgetown-core/test/test_new_command.rb`, but there's nothing testing content pages in the site template specifically. I think we're OK for now…a future goal is to bring an integration testing strategy into the project, either something Rails-like or possibly Cypress, so that might be a good time to test the built content of the site template.

PR #40, opened 2020-05-09
add bundler cache to the build GH action to improve test speed

andrewmcodes:

- [ ] I've added tests (if it's a bug, feature or enhancement)

## Summary

Adds caching for bundled gems to improve test speed on subsequent runs for [GitHub Actions](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows)

I noticed that you are not running the tests or linters for pull requests except on master. I didn't add this in because it would be super quick to add and I figured there may be a reason but I have found [this](https://github.com/gitcoinco/code_fund_ads/blob/master/.github/workflows/standardrb.yml#L3) setup to be ideal as sometimes people will merge stuff into their own branches and tests would not run if they were not pointing at master. 

## Context

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,17 +8,22 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v2
     - name: Set up Ruby 2.6
       uses: actions/setup-ruby@v1
       with:
         ruby-version: 2.6.x
+    - uses: actions/cache@v1
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
     - name: Build and test with Rake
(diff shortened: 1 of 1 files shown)


jaredcwhite:
@andrewmcodes Thanks for this! Anything to speed up the CI builds is great.

The tests/rubocop are running for PRs as well as master, so I'm not sure what you mean there?

andrewmcodes:
So it’s a little hard to explain quickly but I believe with your config, if a user without commit access (aka me) opens a PR to a branch other than master the actions won’t run. Could be wrong bc they have tweaked it a few times but that’s my current understanding. May not be a concern `¯\_(ツ)_/¯`

jaredcwhite:
Ah…so you're saying if the PR is to merge a branch into another *branch*, not master directly, then the action doesn't run? If that's the case, then yes it'd be great to fix that. 👍

jaredcwhite:
LGTM! 👍 

(I'll also add in the fix so all PRs work, not just ones that merge to master)

Issue #25, opened 2020-04-26
Custom plugins don't reload when the site is being watched


jaredcwhite:
This isn't technically a "bug", but I find it extremely annoying that when you are developing a custom plugin in the `plugins` folder, changes to the code don't get picked up while running `build -w` or `serve` and you have to shut down the watcher and reload it. There's got to be a way to do a Rails-style reload automatically. 😊

jaredcwhite:
Fixed in #41

PR #41, opened 2020-05-11
Plugin content and autoreload

### LINKED ISSUE #22: feat: Plugins that can add content to the site (components, layouts, etc.)
This is a "mega" issue to track the ability for Bridgetown plugins to supply a manifest that can be consumed and allow files to be added to the site build from the plugin folders same as if they were in the main site folder. This would include components (aka `{% render "cool_plugin/widget" %}`), layouts, pages, static files, and data. A certain kind of plugin utilizing these features could be called a "theme", but there are plenty of other interesting use cases.

I already have a bunch of this working locally in a proof-of-concept branch, which is pretty neat, but I think some of the usability issues need to be thought through carefully.

* What if you want to override one of the plugin's files? I don't think "fish around in a gem repo and copy the right file" is a good answer. Maybe we need some extra CLI commands.
* What if multiple plugins try writing to the same url/data/etc. location? Should the first one win? The last one? Issue a warning?
* What if the plugin wants to adjust what it provides based on a user's configuration? For that matter, how do we handle configuration? Dump everything in `bridgetown.config.yml`? Use separate data files?
* What if a plugin "goes" rogue and messes up other parts of the site? How would the user know how to troubleshoot the problem? At the very least I think we'll need more debug logging that can get switched on to know exactly which plugin is doing what at which time.
* What if plugin A wants to include plugin B as a dependency? (aka "child themes", or multiple plugin "monorepos")

jaredcwhite:

## Summary

This is the mother of all PRs! :scream:

It adds four marquee features to Bridgetown:

1. Any Ruby files you place in your `plugins` folder will now hot reload whenever you update them or another part of the site as part of the Watcher mechanism. No more stop-and-restart every time you want to test a change to your Liquid tag, or external API parsing, etc.
1. Gem-based plugins can now provide layouts, content (pages or static files), and Liquid components by defining a Source Manifest. This is just a few lines of code which a plugin uses to tell Bridgetown where to look for those files. In a nutshell, _this is the return of Themes!_ Only instead of a site only being able to use one theme, it can use multiple themes in parallel. 🤯 
1. A new command `bridgetown plugins` will let you `plugins list` all custom or gem-based plugins you have loaded in the site along with other related infrastructure such as source manifests, generators, and builders (see below). In addition, you can `plugins cd` directly into a gem folder and use the `BRIDGETOWN_SITE` env var to reference the parent site. So to copy all gem layouts into your own site layouts folder, it's as simple as
  `bridgetown plugins cd AwesomePlugin/layouts`
  `cp -r * $BRIDGETOWN_SITE/src/_layouts`
1. And last but certainly not least…introducing the brand-new Unified Plugins API! 🎉 It's all based around a superclass named `Bridgetown::Builder` and it's a totally different paradigm for writing plugins compared to past experiences with Jekyll. By way of analogy, it's like migrating from Sinatra to Rails.

Quick preview (this is straight from a new plugin gem I'm writing to add Cloudinary image support to Bridgetown sites):

```ruby
module Bridgetown
  module Cloudinary
    class Builder < Bridgetown::Builder
      CONFIG_DEFAULTS = {
        cloudinary: {
          cloud_name: nil,
          default_transformation: "open_graph",
          default_image_format: "jpg",
          transformations: {
            open_graph: "c_fill,g_face:center,w_1600,h_900,q_50",
            tiny: "w_300,c_limit,q_90",
            small: "w_600,c_limit,q_85",
            medium: "w_1200,c_limit,q_80",
            large: "w_1800,c_limit,q_80",
            xlarge: "w_2048,c_limit,q_75",
          },
          add_transformed_urls_to_image_front_matter: false,
        },
      }.freeze

      # Set up the Cloudinary configuration
      def build
        ::Cloudinary.config({
          "cloud_name" => config[:cloudinary][:cloud_name],
          "secure"     => true,
        })

        generator :add_image_urls_to_documents
        liquid_tag "cloudinary_img", :img_tag
        liquid_filter "cloudinary_url", :url_filter
      end

      # Populate front matter
      def add_image_urls_to_documents
        Bridgetown::Cloudinary::Utils.add_image_urls_to_documents(
          site, config[:cloudinary]
        )
      end

      # Define the "cloudinary_img" Liquid tag
      def img_tag(attributes, tag)
        alt, id, transformation = attributes.split(",").map(&:strip)
        alt.delete! '"'
        if id.include? '"'
          id.delete! '"'
        elsif id.include? "."
          obj, var = id.split(".")
          id = tag.context[obj][var]
        else
          id = tag.context[id]
        end
        transformation&.delete! '"'

        "<img alt=\"#{alt}\" src=\"#{url_filter(id, transformation)}\" />"
      end

      # Define the "cloudinary_url" Liquid filter
      def url_filter(id, transformation = nil)
        Bridgetown::Cloudinary::Utils.url(
          config: config[:cloudinary], id: id, transformation: transformation
        )
      end
    end
  end
end

Bridgetown::Cloudinary::Builder.register
```

That, along with a couple of extra utility methods not shown here, is all that is needed to provide a Bridgetown site with a `cloudinary_img` Liquid tag, `cloudinary_url` Liquid filter, and optionally add multiple Cloudinary image urls to the front matter of all pages and documents which specify a `cloudinary_id`. Before the Unified Plugins API, doing that would have required using three completely different low-level APIs (Liquid's native tag and filter APIs along with our Generator API), none of which inherently share object/variable/config scope. (Oh yeah, did I mention you can now define default plugin configuration that users can override in `bridgetown.config.yml`?)

I still have to write documentation for all of these features which will be in a separate PR, but I wanted to get the code/tests out the door here first for general review. Not to overdo the hype here, but I think these features are a **game changer** for the framework.

## Context

Fixes #22, #25 
Partially implements: #31


jaredcwhite:
Oh, I forgot to call out the fact this PR also switches all site/data/front matter hashes to HashWithIndifferentAccess from ActiveSupport. No more driving me bonkers every time I try to `page.data[:title]` 😆 This is but the first of _many_ enhancements we'll be bringing to the framework by leveraging all the goodies provided by AS.

jaredcwhite:
All right, I think we should GTM and address further feedback and potential adjustments in follow-up PRs (first one being the documentation). Let's do this! :shipit:

Issue #24, opened 2020-04-26
feat: Add "automation" to Bridgetown CLI


jaredcwhite:
UPDATE: see comment below — planning on rewriting the CLI in Thor.

----

Similar to how you can do:

```
gatsby new gatsby-material-starter https://github.com/Vagr9K/gatsby-material-starter
```

or 

```
rails new my_app -m https://raw.github.com/webBoxio/rails-templates/master/base_template.rb
```

We'd simply augment the existing `bridgetown new` command to support a path to a folder or a remote URL like GitHub, and then it could use that instead of the built-in site template. Bonus feature would be to support some kind of simple configuration process to fill-in-the-blanks, but adding an interactive Q&A to the command line seems quite complicated, so we may save that for a future enhancement.

jaredcwhite:
While we're testing new starter kit repos, let's set up a `netlify.toml` file too so a new starter site can be deployed to Netlify right away. Something as simple as:

```
[build]
  command = "yarn deploy"
  publish = "output"
```

jaredcwhite:
Bumping this back to 0.15 — want to think things through a little bit more and solicit more feedback. 0.14 will still have plenty of plugin/theme-related yumminess, so that seems reasonable. 😄

jaredcwhite:
Well…this is going to be painful, but we're going to need to switch away from Mercenary (which Jekyll used) and rewrite our CLI in Thor.

Why? So Bridgetown can do things as incredibly cool as this:

https://railsbytes.com/

This is far and away the best example in the Ruby world I've seen (not to mention anywhere else!) of extending new or existing projects for a framework.

It's also worth mentioning that Middleman uses Thor as well. I don't even know if Mercenary is a thing outside of Jekyll.

jaredcwhite:
Done, now available in 0.15.0.beta3 public release.

PR #56, opened 2020-05-19
Migrate CLI from Mercenery to Thor and Enable Automations

jaredcwhite:
As advertised…now using Thor to construct the CLI so that we can use all the extra goodies that affords such as Thor Actions to enable slick automations and interactive commands (see #24).

Besides porting all the existing commands over to Thor, there's now an `apply` command (plus `--apply` for the `new` command) which will run automation scripts which make use of actions. It even can parse GitHub and Gist URLs and look for a `bridgetown.automation.rb` script to execute. So many awesome possibilities — it's like starter kits/themes on steroids.

PR still requires updated documentation.


jaredcwhite:
Merging this into the 0.15 beta release branch. We don't want documentation and other aspects of the release altering the public website yet.

PR #66, opened 2020-06-02
Fix for GitHub branch URLs in automations (0.15)

### LINKED ISSUE #65: Apply command doesn't respect branches on transform
When supplying a branch such as "automation" or "development" to the `bridgetown apply` command, it appends "master"

**Bridgetown Version**:

0.15.0.beta-2

**Current Behavior**:

```bash
bridgetown apply https://github.com/ParamagicDev/bridgetown-plugin-tailwindcss/tree/automation/bridgetown.automation.rb
```

Turns 

"https://github.com/ParamagicDev/bridgetown-plugin-tailwindcss/tree/automation/bridgetown.automation.rb" 

into 

"https://raw.githubusercontent.com/ParamagicDev/bridgetown-plugin-tailwindcss/automation/master/bridgetown.automation.rb"

**Expected Behavior**

"https://github.com/ParamagicDev/bridgetown-plugin-tailwindcss/tree/automation/bridgetown.automation.rb" 

Should turn to

"https://raw.githubusercontent.com/ParamagicDev/bridgetown-plugin-tailwindcss/automation/bridgetown.automation.rb"

ParamagicDev:
This is a 🐛 bug fix.

  - I've added tests (if it's a bug, feature or enhancement)

## Summary

My pull request allow you to use a branch for the command "apply"

It transforms the Github url to one that allows the use of a branch

IE: 

`https://github.com/bridgetownrb/bridgetown/tree/automations`

turns to

`https://raw.githubusercontent.com/bridgetownrb/bridgetown/automations/bridgetown.automation.rb`

## Context

Fixes #65

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
@@ -77,18 +77,22 @@ module Bridgetown
 
       private
 
+      def remote_file(arg)
+        if arg.end_with?(".rb")
+          arg.split("/").yield_self do |segments|
+            arg.sub!(%r!/#{segments.last}$!, "")
+            segments.last
+          end
+        else
+          "bridgetown.automation.rb"
+        end
+      end
+
       # TODO: option to download and confirm remote automation?
       def transform_automation_url(arg)
         return arg unless arg.start_with?("http")
 
-        remote_file = if arg.end_with?(".rb")
-                        arg.split("/").yield_self do |segments|
-                          arg.sub!(%r!/#{segments.last}$!, "")
-                          segments.last
-                        end
-                      else
-                        "bridgetown.automation.rb"
-                      end
+        remote_file = remote_file(arg)
 
         tree_regex = %r!https://github\.com/(?<path>.*/.*)/tree/(?<branch>.*)/?!
         match = tree_regex.match(arg)
--- a/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
@@ -90,10 +90,17 @@ module Bridgetown
                         "bridgetown.automation.rb"
                       end
 
+        tree_regex = %r!https://github\.com/(?<path>.*/.*)/tree/(?<branch>.*)/?!
+        match = tree_regex.match(arg)
+
         if arg.start_with?("https://gist.github.com")
           return arg.sub(
             "https://gist.github.com", "https://gist.githubusercontent.com"
           ) + "/raw/#{remote_file}"
+        elsif match
+          return arg.sub(
+            tree_regex, "https://raw.githubusercontent.com"
+          ) + "/#{match[:path]}/#{match[:branch]}/#{remote_file}"
         elsif arg.start_with?("https://github.com")
           return arg.sub(
             "https://github.com", "https://raw.githubusercontent.com"
--- a/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
@@ -88,6 +88,7 @@ module Bridgetown
         end
       end
 
+
       # TODO: option to download and confirm remote automation?
       def transform_automation_url(arg)
         return arg unless arg.start_with?("http")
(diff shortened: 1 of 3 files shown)


ParamagicDev:
I added a `require bundler` to actions because it was failing locally for me without the require statement.

It said `NameError: uninitialized constant Bridgetown::Commands::Apply::Bundler`

jaredcwhite:
Terrific! I'll dig into this later today 👍

bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
```diff
@@ -77,23 +77,34 @@ def apply_from_url(url)
...
 
-        remote_file = if arg.end_with?(".rb")
-                        arg.split("/").yield_self do |segments|
-                          arg.sub!(%r!/#{segments.last}$!, "")
-                          segments.last
-                        end
-                      else
-                        "bridgetown.automation.rb"
-                      end
+        remote_file = remote_file(arg)
```
jaredcwhite, line 95: I generally prefer to name local variables differently than methods so there's no confusion. Not sure what makes sense here…would love suggestions.

bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
```diff
@@ -77,23 +77,34 @@ def apply_from_url(url)
...
 
         if arg.start_with?("https://gist.github.com")
           return arg.sub(
             "https://gist.github.com", "https://gist.githubusercontent.com"
           ) + "/raw/#{remote_file}"
+        elsif match
+          return arg.sub(
+            tree_regex, "https://raw.githubusercontent.com"
+          ) + "/#{match[:path]}/#{match[:branch]}/#{remote_file}"
         elsif arg.start_with?("https://github.com")
```
jaredcwhite, line 108: Do we need still need to check GitHub with `start_with?` if we're doing the match above?

bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
ParamagicDev: Honestly, I only did this to make rubocop happy. Could name it something like `find_remote_filename(arg)` and then do 

`remote_file = find_remote_filename(arg)`

jaredcwhite: How about `determine_remote_filename`

bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
ParamagicDev: The match above wont match with a regular master branch because it doesnt include `tree`.

However, we could make a base regex for github:

IE:

```ruby
github_regex = %r!https://github\.com!
```

And then use Regex concatenation to drop it into the tree regex like so:

```ruby
github_tree_regex = %r!#{github_regex}/(?<path>.*/.*)/tree/(?<branch>.*)/?!
```

First we want to run the matches up front before we mess with the URL.
Then we will always replace Github on github_regex match
Something like the following:

```ruby
github_match = github_regex.match(arg)
github_tree_match = github_tree_regex.match(arg)

elsif github_match
  new_url = arg.sub(github_regex, "https://raw.githubusercontent.com")
  
  return new_url + "/master/#{remote_file}" if github_tree_match.nil?

  return new_url + "/#{match[:path]}/#{match[:branch]}/#{remote_file}"
end
```

ParamagicDev: On further review, we actually dont need the `match[:path]` we could simple do `#{match[:branch}/#{remote_file}`

jaredcwhite: That sounds like a good approach to me. Can you go ahead and update your branch with the conditional logic consolidated like that? 👍

bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
ParamagicDev: sounds good, ill make it happen.

bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
ParamagicDev: Absolutely, working on it now, the issue I'm running into is I have to sub "/tree/" but I dont want to run the risk of having the user have a directory named `/tree/` so it may be slightly more complicated than expected. Ill add a test for making sure you dont conflic with a `/tree/` directory.

jaredcwhite:
Wow, the logic for this is badass…took me a while even to grok how it was working, so bravo. 👏 Merging!

Issue #48, opened 2020-05-14
feat: Integration tests (Cypress? Something from Rails? etc.)


jaredcwhite:
We should integrate and demonstrate how to write integration tests that can be run in CI or during development to verify various aspects of site output, routes, components, etc. The field is wide open as far as I'm concerned as to the best way to do that. If we can stick to something simple and Ruby-based, that's cool. But I'm also willing to go for something heavy-duty like Cypress if that will better serve the ecosystem in the long run.

jaredcwhite:
The 0.15 release is pretty loaded, but as a stretch goal I think at least some foundational element of this could be included, perhaps as a blank folder in the site template or a suggested approach in the documentation.

ParamagicDev:
Ugh. Testing terminology. 

When I hear `Integration Test` I think something that parses everything under the hood and doesn't use a browser. 

IE:

[DOM Testing Library](https://testing-library.com/docs/dom-testing-library/intro)

[Enzyme](https://github.com/enzymejs/enzyme)

Then there's End to End testing which would be something like:

[Cypress](https://www.cypress.io/)

It seems like [Capybara](https://github.com/teamcapybara/capybara) weirdly hits both aspects based on my limited knowledge.

It can hook into Minitest or RSpec and work as integration testing. But it can also work as a Selenium driver (end to end testing) if I'm reading the docs correctly.

Most of the Rails / Ruby community would probably prefer Capybara due to familiarity.

But most of the front end world uses something like Enzyme / React-Testing-Library for integration tests and Cypress for end to end testing.

Sorry, just thinking out loud here. I'd be happy to see if I could throw something together with either Cypress and Dom-testing-Library or Capybara.

 Theoretically they're both framework agnostic and shouldn't require anything special to get working.

jaredcwhite:
I started to go down a path of using Minitest + Nokogiri via a Bridgetown hook to test the literal HTML at the point it's been statically generated, and so far I'm liking that approach as one possibility. (see #69)

If you want to give Cypress a spin for E2E testing, I think that'd be awesome. I've used Cypress a bit and it's pretty nifty, but I think it's mostly relevant only for parts of the site that are interactive and require JavaScript.

ParamagicDev:
I just remembered Rails has a DOM testing library. It uses Minitest / Nokogiri under the hood.

[Rails DOM Testing Library](https://github.com/rails/rails-dom-testing)

I wonder if there would be any way to leverage those assertions.

[...]

PR #69, opened 2020-06-03
Website experiment with test suite

jaredcwhite:
Adding a Minitest/Nokogiri test suite via a `post_write` hook to the Bridgetown website. Initial commit here should fail, as I want to test how Netlify handles the build. It should give us a red X, and if I fix the test we'll get a green checkmark.

(This is in reference to #48)


ParamagicDev:
Don't want to make extra work for you. I just remembered Rails has a DOM testing library. It uses Minitest / Nokogiri under the hood.

[Rails DOM Testing Library](https://github.com/rails/rails-dom-testing)

I wonder if there would be any way to leverage those assertions.

jaredcwhite:
Started a page about Automated Testing:
https://deploy-preview-69--bridgetownrb.netlify.app/docs/testing

PR #76, opened 2020-06-12
Use AwesomePrint gem for console

jaredcwhite:
AwesomePrint provides a great experience in the console for displaying data, so let's use it!

Also refactored the GitHub regex a bit in the Apply command and brought back the --no-watch option for the Serve command.


jaredcwhite:
LGTM! :shipit:

PR #79, opened 2020-06-19
ERB/Slim/Haml template support

jaredcwhite:
_WIP – do not merge! (see comments at bottom)_

So in a nutshell, Tilt is (unsurprisingly I suppose) quite awesome, and it means we can support a variety of new template languages with minimal fuss. This PR facilitates loading both page/document and layout templates ending in `.erb`, `.haml`, and `.slim` — all supporting front matter just like Liquid templates — and a number of Bridgetown objects will be provided as well. I also added support for rendering Liquid components via `liquid_render` as well as a template-specific `partial_render` (which will look in `src/_partials`) so you can render ERB partials from ERB templates, Slim partials from Slim templates, etc.

For example, here's an ERB page:

```erb
---
title: Whoa, this is amazing!
---

<h1><%= page.data[:title] %></h1>

<%= liquid_render "some_other_component", page: page, site: site_drop %>

<%= partial_render "cool/stuff", hehe: "har har" %>

<p>I am done!</p>
```

And here's a Slim layout:

```slim
---
super_groovy: 123
---
doctype html
html lang="en"
  head
    == liquid_render "shared/head", metadata: site.metadata, title: page.data[:title]
  body class="#{page.data[:layout]} #{page.data[:page_class]}"
    == liquid_render "shared/navbar", page: page, version: Bridgetown::VERSION, metadata: site.metadata

    | Super groovy!
    == layout.data[:super_groovy]

    == yield

    == partial_render "very_slim", hehe: "woo hoo"

    | How many collections?
    == site.collections.count
```

**However** — I'm inclined to leave all the Tilt converters out of core and make them separate plugins. I'm not yet convinced this should all be in core. I think having `RubyTemplateView` and the other rendering pipeline tweaks in core makes sense, and then each individual template converter could be its own gem (`bridgetown-erb`, `bridgetown-slim`, etc.). Or we could have one `bridgetown-tilt` gem which bundles ERB, Haml, and Slim for anyone to mix 'n' match.

What do you think? (Also any feedback on the component/partial handling? Any new helpers we might want to throw in the mix?)


ParamagicDev:
My opinion is that `tilt` should remain in core as well as 1 templating language. My personal preference would be ERB simply because it comes bundled with Ruby by default and its how Rails handles templating.

Also, you're not stuck supporting multiple templating languages within the core stack and adding multiple runtime dependencies like `slim`, `haml`, etc.

If anyone wants to support additional templating languages, its as simple as creating a gem, use `bridgetown-core` as a dependency, inherit the `RubyTemplateView` class, and add a `Converter`.

Also, the render functions are pretty slick and are a nice way of handling liquid components. Overall, it looks super solid!

jaredcwhite:
Cool, that does make sense to me now. All right, off to modularize and write up some tests, docs, etc.

jaredcwhite:
Oh, one thing I forgot about — there's no obvious path towards doing the kind of multi-extension stuff you find in Rails or Middleman — aka, `filename.html.erb` vs. `filename.json.erb`. Right now everything's defaulting to rendering to `.html`. I wonder if that's a showstopper at this point or if it can be a "V2" feature down the road…

jaredcwhite:
Been a bit distracted lately but I'm wrapping up the initial version of this with ERB support and then plan to release separate plugins for Haml and Slim separately. Should all be pretty straightforward in usage which is neat. I'll be curious how this feature tickles the fancy of folks coming from Rails and even Middleman…

jaredcwhite:
OK, I think I'm done tinkering. Wrote up a bunch of tests. Now I just need to write docs. BTW, I renamed `partial_render` to just `partial` to match Middleman, and changed it so the partial filename has to start with `_` like other Ruby frameworks.  I also integrated in the same filters as are provided to Liquid so you can call them like standard ERB helpers which is pretty slick.

@ParamagicDev Can I ask for a second pair of eyes before merging this in?

bridgetown-core/lib/bridgetown-core/renderer.rb
```diff
@@ -140,9 +141,13 @@ def render_liquid(content, payload, liquid_context, path = nil)
     # Convert the document using the converters which match this renderer's document.
     #
     # Returns String the converted content.
-    def convert(content)
+    def convert(content, document)
       converters.reduce(content) do |output, converter|
-        converter.convert output
+        if converter.method(:convert).arity == 1
```
ParamagicDev, line 146: Pretty slick way of checking if document is defined or not :+1:

PR #84, opened 2020-07-01
Reducing animation for users who prefer reduced motion

MikeRogers0:

## Summary

The animations when navigating between pages on [the documentation](https://www.bridgetownrb.com/docs/core-concepts) are giving me a headache.

image

I have reduced motion enabled so expect less animation. This removes the animation for users who have "Reduce Motion" checked.

## Context

I've not opened an issue, I created a PR to experiment with solution & get feedback.

### DIFF
--- a/bridgetown-website/frontend/javascript/index.js
+++ b/bridgetown-website/frontend/javascript/index.js
@@ -52,19 +52,21 @@ document.addEventListener('DOMContentLoaded', () => {
     containers = [mainEl, "#topnav"]
   }
 
-  const swup = new Swup({
-    containers: containers,
-    plugins: [
-      new SwupSlideTheme({mainElement: mainEl}),
-      new SwupBodyClassPlugin(),
-      new SwupScrollPlugin({animateScroll: false})
-    ],
-    
-  })
+  if(!window.matchMedia('(prefers-reduced-motion)')) {
+    const swup = new Swup({
+      containers: containers,
+      plugins: [
+        new SwupSlideTheme({mainElement: mainEl}),
+        new SwupBodyClassPlugin(),
+        new SwupScrollPlugin({animateScroll: false})
+      ],
 
-  swup.on('contentReplaced', function() {
-    addHeadingAnchors()
-  })
+    })
+
+    swup.on('contentReplaced', function() {
+      addHeadingAnchors()
+    })
+  }
 
   addHeadingAnchors()
 })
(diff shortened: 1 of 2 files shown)


ParamagicDev:
I see no issues with this. Seems to have pretty wide browser support.

[https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)

Also, you beat me to the review, but the `@media` query needs to be outside of the class. You fixed it already though :)

 I see no reason why a user cant prefer reduced motion. As long as @jaredcwhite is okay with it, I dont see any reason why this cant be merged.

MikeRogers0:
@ParamagicDev  It looks like the media query is being stripped at some point during the build :/ Any idea of how to fix that?

bridgetown-website/frontend/javascript/index.js
```diff
@@ -52,19 +52,21 @@ document.addEventListener('DOMContentLoaded', () => {
...
-  const swup = new Swup({
-    containers: containers,
-    plugins: [
-      new SwupSlideTheme({mainElement: mainEl}),
-      new SwupBodyClassPlugin(),
-      new SwupScrollPlugin({animateScroll: false})
-    ],
-    
-  })
+  if(!window.matchMedia('(prefers-reduced-motion)')) {
```
jaredcwhite, line 55: I think you need a .matches on the end there to account for cases where it's false — right now this seems to be true all the time.

ParamagicDev: @jaredcwhite appears to be correct. As of now it's only testing that the property exists, but doesn't actually test whether or not the property value is true / false.

Just figured I'd also drop the relevant document on `matchMedia` just for future reference in case we ever revisit this.

[https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Testing_media_queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Testing_media_queries)

Thanks again for the PR :)

MikeRogers0, line 55:
```suggestion
  if(!window.matchMedia('(prefers-reduced-motion)').matches) {
```

MikeRogers0: Ah my mistake (classic lunchtime coding!). All sorted :)

Thank you for your help (and awesome responsiveness!) on this PR :)

Issue #91, opened 2020-07-05
Investigate adding Zeitwerk to core gems


jaredcwhite:
Because we are pulling in ActiveSupport now, it's bringing the [Zeitwerk gem](https://github.com/fxn/zeitwerk) along for the ride. If we can leverage that to make the process of requiring or autoloading parts of Bridgetown easier and/or faster, that would be groovy.

ParamagicDev:
Based on the current file structure, this shouldnt be too difficult to implement. Based on what I see from the module / file structure of Bridgetown, we may just have to throw in a few `loader.collapse('*/<directory>')`.

IE:

`loader.collapse(''*/concerns") # to avoid making a `Concerns` module.`

[https://github.com/fxn/zeitwerk#collapsing-directories](https://github.com/fxn/zeitwerk#collapsing-directories)
I can take a look into this at some point and see if I can implement it and not break any tests. Ill go ahead and assign myself to this.

ParamagicDev:
@jaredcwhite should we consider explicitly adding `Zeitwerk` Rather than relying on an implicit dependency on the off-chance active-support drops it.

jaredcwhite:
@ParamagicDev Yes. ☺️

jaredcwhite:
This might help streamline some aspects of this:

https://twitter.com/fxn/status/1283166824100761600?s=21

PR #95, opened 2020-07-08
feature/zeitwerk

ParamagicDev:
WIP: Do not Merge

In reference to #91 

Barebones implementation of Zeitwerk. Cant say I'm totally happy with it. 

It was not exactly as plug and play as I thought it would be. Ran into issues with undefined constants, still needing to use require calls internally, etc. This is working, but really didnt change the original code. Feel free to play around with it.

Honestly, not even totally sure Zeitwerk is doing anything here...


jaredcwhite:
This looks really interesting. It might not be "doing" much at the moment with all the other requires/autoload statements in there, so probably we'd need to strip all those out and then see if Zeitwerk alone can help resolve everything. Also, in theory, a site build would go faster if it doesn't have to eager load/require every file in the Bridgetown gem each time, but only the ones that happen to get used for a particular site… I'm not sure if that's really feasible though, aka, there might be very few Ruby code files that are "optional" for most site builds.

ParamagicDev:
@jaredcwhite I tried stripping out the requires and autoloads and it was an exercise in hair pulling for sure. But I do agree it could have bigger implications for the future.

jaredcwhite:
I've been waffling back and forth on this, but I'm inclined to want to keep pushing this forward. I think nothing but good things are in store if Bridgetown continues to take advantage of the Ruby ecosystem we have today courtesy of Rails. I'll try to set aside some time soon to fiddle with this…

andrewmcodes:
+1 to @jaredcwhite's point about pushing the community forward. My utmost respect to you for diving into this. 

I have spent the last few days in a cage match with autoloading and zeitwerk and it's driving me batty.

jaredcwhite:
I think I will close this only because we've incrementally added Zeitwerk in more recently in certain cases and particularly in PR #243 so it's probably more of a general refactor over time than a single switch. Thanks for your work on this @ParamagicDev to kick the tires though!

PR #98, opened 2020-07-11
Update pagination documentation

andrewmcodes:

## Summary

Update pagination documentation to include information on how to use the `per_page` key.

## Context

I do not currently see a place in the docs that we call out how to set the amount of paginated items per page. I had to go look in the source code, so this addition adds a little blurb about the `per_page` key to the pagination docs to make it easier to find.


jaredcwhite:
@andrewmcodes Thanks! There are a few other options that still need to get documented, but this is a good start 😃￼

Issue #86, opened 2020-07-03
Class map for Liquid components


jaredcwhite:
By far the messiest code pattern I've seen while working on Liquid components and templates in general is constructing a string of CSS class names based on incoming variables. Typically in frontend component libraries you'll see a "class map" function that makes this process pretty straightforward. Prior art:

* https://lit-html.polymer-project.org/guide/styling-templates
* https://github.com/JedWatson/classnames

It would be great to have a Liquid tag to make this easy. Something like:

```
<button class="{% class_map is-small: small, has-text-center: centered, outlined: !filled %}">…</button>
```

We'd have to do some fancy parsing to load variables from local scope and work with conditional values, but the actual business of constructing the output string should be pretty easy.

I'm also tempted to make this its own little gem so it can be used anywhere, but maybe we can implement here first and extract it out at some point once it's working well.

ParamagicDev:
So basically class_map takes in a hash, the key is the class, and value is Boolean. If the Boolean is true, then output the key, if the Boolean is false, ignore the key.

Rough example:
  Input:

```
<button class="{% class_map text-center: true, bg-red-500: "string", shadow: false %}">I'm a button</button>
```
 
 Output:

```
<button class="text-center bg-red-500">I'm a button</button>
```

jaredcwhite:
Added in #99

PR #99, opened 2020-07-12
feat(tags): Adding a class_map tag

ParamagicDev:

Rough implementation, appears to pass normal use cases. Due to how this is implemented, only variables can be given as values. If the variable is false, nil, etc, it will not display in the output. Let me know if there's any questions.

## Summary

Adds a `{% class_map %}` tag as discussed in #86. 

Adds testing for the `class_map` tag

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/tags/class_map.rb
+++ b/bridgetown-core/lib/bridgetown-core/tags/class_map.rb
@@ -39,12 +39,12 @@ module Bridgetown
       end
 
       def render(context)
-        to_class_ary(@input, context).join(" ")
+        class_map(@input, context)
       end
 
       private
 
-      def to_class_ary(string, context)
+      def class_map(string, context)
         ary = []
 
         string.split(%r!,\s+!).each do |item|
@@ -67,7 +67,9 @@ module Bridgetown
           end
         end
 
-        ary
+        ary.join(" ")
+      rescue NoMethodError
+        "invalid-class-map"
       end
 
       def find_variable(context, variable)
--- a/bridgetown-core/lib/bridgetown-core/tags/class_map.rb
+++ b/bridgetown-core/lib/bridgetown-core/tags/class_map.rb
@@ -1,52 +1,75 @@
 # frozen_string_literal: true
-require 'json'
+require 'set'
 
 module Bridgetown
   module Tags
+
+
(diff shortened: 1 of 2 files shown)


bridgetown-core/lib/bridgetown-core/tags/class_map.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        ary.join(" ")
+
+      # Gracefully handle if syntax is improper
+      rescue NoMethodError
+        "invalid-class-map"
+      end
+
+      def find_variable(context, variable)
```
jaredcwhite, line 77: Can we use the `lookup_variable` method from the `Bridgetown::LiquidExtensions` mixin? (TBH, I didn't really notice that was there until very recently! Should probably relocate the file to the `concerns` folder)

ParamagicDev: I tried, but the issue is that it will return the string of the variable instead of nil. I need it to return nil so it can be omitted from output.

IE:

If the variables `foo` is nil, it will return `"foo"`

ParamagicDev: [https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/bridgetown-core/liquid_extensions.rb#L5-L20](https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/bridgetown-core/liquid_extensions.rb#L5-L20)

jaredcwhite: You know, I saw that `lookup || variable` bit and it threw me for a loop at first. I'm not even sure if that's desirable over the long term. Funny thing — I did a search and neither Bridgetown nor Jekyll seems to be using this in any production code. It seems it was added back in 2014 and nobody used it (maybe some third-party Jekyll plugin is using it, I dunno). I feel we could probably change the behavior to return nil…Liquid handles nil gracefully anyway and I simply can't imagine a case where anyone would *want* the variable name returned as a string.

jaredcwhite:
I'll go ahead and merge this in order to get the 0.16 beta gem out the door…will follow up on documentation and simplifying that Liquid var lookup…

PR #101, opened 2020-07-13
Provide a Liquid find tag as easier alternative to where_exp

### LINKED ISSUE #90: feat: new Liquid tag "find"
## Summary

Jekyll, and hence Bridgetown, have had `where` and `where_exp` Liquid filters for quite some time now to query collections and data sets. Due to demand for a way to pluck out a single result more easily, Jekyll recently added `find` and `find_exp` filters. ([reference](https://jekyllrb.com/docs/liquid/filters/))

I'll be honest, I've never been a huge fan of using filters to construct queries. I think using a dedicated tag looks cleaner, puts assigning variables and querying collections into one operation, and makes it easier to remember.

I propose a single new tag: `find`. In theory, this one tag could replace _all four filters_ from Jekyll in usage.

## Explanation
Here's an example of the proposed tag syntax (single object return):

```
{% find album in site.albums, band == page.band, year >= 1980, categories contains "Rock" %}
```

This would look in the `albums` collection, look for an item where the band front matter variable equals the current page's band variable, the year front matter variable is greater than or equal to 1980, and the categories front matter variable contains an entry matching "Rock" — and if found, assign that item to the local `album` variable.

To return an array instead of a single object, the above could simply be written:

```
{% find albums where site.albums, band == page.band, year >= 1980, categories contains "Rock" %}
```

## Drawbacks

The one benefit of filters is they can be chained together, so you might have a `where`-style filter composed with a few other filters. However, in actual usage I almost never see a lot of filters used in conjunction with where/where_exp specifically. And you can still use filters in later assignments:

```
{% find github_authors where site.authors, github: true %}
{% assign sorted_authors = github_authors | sort: "name" %} 
```

So I don't think that's a huge concern.

I'm also curious what the performance implication for using this tag vs using the previous filter.

## Syntax Comparison vs. `where_exp`

```
old and busted:

{% assign movies = site.movies | where_exp: "item", "item.genre == 'horror' and item.language == 'English'" %}

new hotness:

{% find movies where site.movies, genre == "horror", language == "English" %}
```

## Unresolved Questions

How to parse all the conditions in the tag… 🤔

I'm guessing it'll be a combination of leveraging internal Liquid magic and reusing code from the `where_exp` filter. Not insurmountable, but also a fairly involved task which will also need a lot of tests to make sure we don't end up with broken queries. In the end I feel it will be worth it though.

Thoughts? Ideas? Questions?

jaredcwhite:
Closes #90

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/tags/find.rb
+++ b/bridgetown-core/lib/bridgetown-core/tags/find.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module Bridgetown
+  module Tags
+    class Find < Liquid::Tag
+      include Bridgetown::Filters::ConditionHelpers
+      include Bridgetown::LiquidExtensions
+
+      SYNTAX = %r!^(.*?) (where|in) (.*?),(.*)$!.freeze
+
+      def initialize(tag_name, markup, tokens)
+        super
+        if markup.strip =~ SYNTAX
+          @new_var_name = Regexp.last_match(1).strip
+          @single_or_group = Regexp.last_match(2)
+          @arr_name = Regexp.last_match(3).strip
+          @conditions = Regexp.last_match(4).strip
+        else
+          raise SyntaxError, <<~MSG
+            Syntax Error in tag 'find' while parsing the following markup:
+
+            #{markup}
+
+            Valid syntax: find <varname> where|in <array>, <condition(s)>
+          MSG
+        end
+      end
+
+      def render(context)
+        @group = lookup_variable(context, @arr_name)
+        return "" unless @group.respond_to?(:select)
+
+        @group = @group.values if @group.is_a?(Hash)
+
+        expression = @conditions.split(",").map do |condition|
+          "__find_tag_item__.#{condition}"
+        end.join(" and ")
+        @liquid_condition = parse_condition(expression)
+
+        context[@new_var_name] = if @single_or_group == "where"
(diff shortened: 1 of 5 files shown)


bridgetown-core/lib/bridgetown-core/tags/find.rb
```diff
@@ -0,0 +1,72 @@
...
+        end
+      end
+
+      def render(context)
+        @group = lookup_variable(context, @arr_name)
+        return "" unless @group.respond_to?(:select)
+
+        @group = @group.values if @group.is_a?(Hash)
+
+        expression = @conditions.split(",").map do |condition|
```
jaredcwhite, line 35: Hmm, I think I'm being naive here…if somebody has a comma inside of a string value, this is going to mess this up. I probably need to do a more intelligent syntax scan. Besides, I want to implement "or" logic at some point as well which will need further parsing smarts.

jaredcwhite:
Just adding a note here I was hoping to get this done in time for final 0.16 release but it looks like it'll have to get pushed back to the next release.

jaredcwhite:
Circling back around on this…I added a very simplistic string scan to convert commas outside of double-quoted strings into a special separator that will be used for the condition split. There's a much better way to do this but I think this is fine for a v1.

jaredcwhite:
I'm very pleased with the bit of Liquid syntax on that website that can now be:

```
{% find next_page in site.docs, order == next_order %}
```

PR #106, opened 2020-07-18
Use .liquid as the default extension for layout files.

wout:

## Summary
Changes extensions on layout files from `.html` to `.liquid`


wout:
Not sure why this fails. It passes locally when running either `rake` or `cd bridgetown-core && script/cibuild`.

jaredcwhite:
I ran it again and the tests passed. I saw that happen once before lately. I think a nondeterministic failure point has crept into the suite. 😕

jaredcwhite:
Thanks @wout — I think this will help make layouts and components feel more akin to each other, and also bring it inline with the .erb convention for ERB templates. I also updated the website repository itself to have .liquid layouts.

PR #107, opened 2020-07-22
Replace 'insert_svg' with 'bridgetown-inline-svg' plugin on website

andrewmcodes:

## Summary

- Adds the [bridgetown-inline-svg](https://github.com/andrewmcodes/bridgetown-inline-svg) plugin for easier handling and inline of SVG assets on the website.
- Removes `insert_svg` tag that was created to handle this previously

## Context

@jaredcwhite mentioned he wanted to do this so I decided to save him the work! 💪

### DIFF
--- a/bridgetown-website/Gemfile
+++ b/bridgetown-website/Gemfile
@@ -5,9 +5,12 @@ gem 'bridgetown-builder', path: '../bridgetown-builder'
 gem 'bridgetown-paginate', path: '../bridgetown-paginate'
 gem 'bridgetown', path: '../bridgetown'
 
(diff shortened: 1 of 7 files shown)


jaredcwhite:
Thanks @andrewmcodes, looks good. One little thing I noticed is the previous tag was stripping out newlines so the entire SVG gets printed out on a single line, whereas this plugin is preserving the newlines. Not a huge deal, but I think if the plugin is doing an optimization step it should probably strip out the newlines as well. Regardless, I'll go ahead and merge. 👍

Issue #103, opened 2020-07-16
Improve Lighthouse scores on website


jaredcwhite:
_Update: mostly improved now, see 7/27 comment below_

While I'm not inclined to go nuts getting to 100 on all the scores, I *was* pretty happy with 90+ numbers when the Bridgetown site first launched, and now our scores suck in large part because I was a bozo and added a YouTube video embed directly on the homepage.

Anyway, this is a general issue to take another look and see if we can improve things. I also noticed the Inter font files are pretty large so maybe we can optimize that as well.

andrewmcodes:
Peaked my interest so I took a look and I could bump most of these but there will be tradeoffs to the design for some that you'll have to decide the importance of. The font is definitely the killer for performance. Lots of interesting research out there on how to get some gains here.

### Desktop

![image](https://user-images.githubusercontent.com/18423853/88144625-3cf76c80-cbc7-11ea-9316-e6592aee2cc6.png)

### Mobile

![image](https://user-images.githubusercontent.com/18423853/88144694-5698b400-cbc7-11ea-88e9-31155b40ca3a.png)

## Quick Wins

Best practices would be easy to get to 100. `Add 'rel="noopener"' or 'rel="noreferrer"' to any external links to improve performance and prevent security vulnerabilities.`

SEO to 100 seems easy enough without compromising the style. Its upset about a few links in the footer which doesn't make a ton of sense to me but w/e. 

Accessibility will be the hardest imo - if you aren't building for it out of the gate, it can be hard to add in retroactively. Especially the sequential ordering of heading elements. Font seems to be the biggest drain on the perf which wouldn't be hard to solve but may be hard to do well. 

Lmk if you want to talk more about this! Web perf is definitely interesting to me. Ping me on discord bc I am a bit slow to respond on GH :shrug:

MikeRogers0:
I opened up a few PRs so you can pick and choose the performance changes you like 👍 

For the fonts:

1. It looks like you're pulling all of fonts from https://github.com/ForkAwesome/Fork-Awesome (~100kb or so) but only using a handful of them. I'd suggest https://feathericons.com/ as an alternative & going SVG for all of them if that's something you're interested in?
2. The `Inter` font is pretty heavy, do you know why you picked it?

jaredcwhite:
Thanks @andrewmcodes & @MikeRogers0 

1. Def interested in streamlining icon usage, I don't think we're actually using very many anyway.
[...]

PR #108, opened 2020-07-22
feat: Lazy loading iframes & adding title attribute to them

MikeRogers0:

## Summary

This add the [`loading="lazy"`](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#Images_and_iframes) method to the iframes, which means the browser won't load the iframe until the viewport is close to showing that iframe. It also adds a `title` attribute to improve accessibility.

## Context

I measured via https://web.dev/measure/ 

**Current Lighthouse**

image

**Review App Lighthouse**

image

As you can see it's a minor improvement, though hard to know if it's just lighthouse hitting a different URL & caching differently (The SEO change is due to the review apps being no-index).

**What's happening?**

image

If you look kind of close at the network order, it's loading the YouTube iFrame after the page has loaded (after the fonts).

## Related Tickets

https://github.com/bridgetownrb/bridgetown/issues/103


jaredcwhite:
@MikeRogers0 Excellent! Thanks for the optimization

PR #113, opened 2020-07-23
Adding loading="lazy" to images near footer

MikeRogers0:

## Summary

This lazy loads of images which are not in the initial viewport of the homepage. It's the same approach I used on the iFrame & it looks super effective on https://web.dev/measure/

Only catch with this PR is you can see the images loading in if you scroll fast enough.

## Context

Continuing on from https://github.com/bridgetownrb/bridgetown/pull/108 which was inspired by https://github.com/bridgetownrb/bridgetown/issues/103


jaredcwhite:
Looks good to me! :+1:

PR #124, opened 2020-08-15
Add Dokku installation instructions to the guides section in the website

wout:

## Summary
This PR adds instructions to the deployment section in the guides on how to configure your Bridgetown project to deploy it to a Dokku server. The same instructions may work as well on Heroku, but I haven't tested it.

The guide is a bit longer than the rest of the instructions on that page and I wasn't sure if it needed a separate page or not. Please let me know if I need to change the wording/writing style or put it in a separate page.


jaredcwhite:
Thanks @wout! I think this is fine. There a bunch of stuff we're hoping to include on the Deployments page going forward and at some point maybe it makes sense to break it up, but no need at present.

Issue #50, opened 2020-05-14
Simplify front matter defaults syntax


jaredcwhite:
**Update:** we ended up opting for a "data cascade" folder-based approach rather than changing the syntax of the defaults key in the config. See #139 

----

I've always found the YAML syntax for defining front matter defaults to be extremely obtuse and I _always_ have to look up a reference to remember how the heck to write it out. Really all I want to do is:

```yaml
defaults:
  posts:
    layout: post
  pages:
    layout: page
  "some/path/to/stuff":
    category: stuff
```

Let's make this a whole lot better and then deprecate the old syntax with the goal of removing it for the 1.0 release.

Note: it should be a pretty straightforward swap because the old syntax is array-based, whereas the new one will be hash-based, so we can easily detect which syntax is being used.

jaredcwhite:
We should do a full performance review of front matter defaults in general as part of this issue. Related issue in Jekyll: https://github.com/jekyll/jekyll/pull/8188/files (it's unclear to me why lazily-loading defaults improves the build time so much…not something I would expect)

ParamagicDev:
So heres the current syntax from `bridgetown-website/bridgetown.config.yaml` [https://github.com/bridgetownrb/bridgetown/blob/8c0132909d3ca5cfe043df37865aeafb2fd96280/bridgetown-website/bridgetown.config.yml#L17](https://github.com/bridgetownrb/bridgetown/blob/8c0132909d3ca5cfe043df37865aeafb2fd96280/bridgetown-website/bridgetown.config.yml#L17)

The `- scope:` syntax is pretty gross I agree and I do like the idea of specifying the path as the key value.

I do like the idea of it being top level in the config rather than being nested into a `_data` folder.

However, I have also thought about the idea of the user defining defaults inside of a `defaults.yaml` file at the top level of a directory.

IE:

```
# ./src/_posts/defaults.yaml

layout: post
author: "Jared White"
```

jaredcwhite:
@ParamagicDev I agree performance should be a concern, but since the number of defaults files would probably be fairly low for an entire site even with thousands of individual pages, I can't imagine it would be a huge hit over a single file. I really like your idea of a "data cascade" based on the filesystem itself — both Eleventy and Hugo have similar concepts, for example.

jaredcwhite:
Update: I got it working with subfolder trees out of the gate. 😃 See #139

jaredcwhite:
[...]

PR #139, opened 2020-09-08
Process data cascade for folder-based frontmatter defaults

jaredcwhite:
Related to #50 

This PR provides the ability to put a `_defaults.yml/yaml/json` file(s) anywhere in the source tree, which will then cause a "data cascade" — aka pages/documents will pull their front matter defaults from one or more defaults files in the tree with subfolders potentially overriding values from their parents.

While this isn't a 100% replacement for the previous defaults config syntax, as you can see here with the changes made to the Bridgetown website, it very well can be a wholesale replacement.

Still needs documentation/tests.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/frontmatter_defaults.rb
+++ b/bridgetown-core/lib/bridgetown-core/frontmatter_defaults.rb
@@ -54,6 +54,10 @@ module Bridgetown
     #
     # Returns the default value or nil if none was found
     def find(path, type, setting)
+      merged_data = {}
+      merge_data_cascade_for_path(path, merged_data)
+      return merged_data[setting] if merged_data.key?(setting)
+
       value = nil
       old_scope = nil
 
@@ -74,6 +78,9 @@ module Bridgetown
     # Returns a hash with all default values (an empty hash if there are none)
     def all(path, type)
       defaults = {}
+
+      merge_data_cascade_for_path(path, defaults)
+
       old_scope = nil
       matching_sets(path, type).each do |set|
         if has_precedence?(old_scope, set["scope"])
@@ -87,6 +94,16 @@ module Bridgetown
     end
 
     private
+
+    def merge_data_cascade_for_path(path, merged_data)
+      absolute_path = @site.in_source_dir(path)
+      @site.defaults_reader.path_defaults
+        .select { |k, _v| absolute_path.include? k }
+        .sort_by { |k, _v| k.length }
+        .each do |defaults|
+          merged_data.merge!(defaults[1])
+        end
+    end
 
     # Checks if a given default setting scope matches the given path and type
     #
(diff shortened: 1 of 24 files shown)


bridgetown-core/lib/bridgetown-core/frontmatter_defaults.rb
```diff
@@ -88,6 +95,16 @@ def all(path, type)
...
 
+    def merge_data_cascade_for_path(path, merged_data)
+      absolute_path = @site.in_source_dir(path)
+      @site.defaults_reader.path_defaults
+        .select { |k, _v| absolute_path.include? k }
+        .sort_by { |k, _v| k.length }
+        .each do |defaults|
+          merged_data.merge!(defaults[1])
+        end
+    end
```
ParamagicDev, line 106: So we basically first check that it exists, then we sort by length because in theory a cascade would look like this:

`src/_defaults.yml`, `src/_posts/_defaults.yml` so the file path will always be longer the further nested you go, pretty cool idea! and then you jsut merge in the defaults array. This is so slick and elegant.

jaredcwhite: Right, it finds any matching parent paths (could be one, could be several…or none), and then does a progressive merge through the paths from highest level to lowest level. Getting those levels via string length sort is perhaps naive, but I can't think of a reason why that wouldn't work.

bridgetown-core/lib/bridgetown-core/readers/defaults_reader.rb
```diff
@@ -0,0 +1,25 @@
...
+module Bridgetown
+  class DefaultsReader
+    attr_reader :site, :path_defaults
+
+    def initialize(site)
+      @site = site
+      @path_defaults = ActiveSupport::HashWithIndifferentAccess.new
+    end
+
+    def read
```
ParamagicDev, line 12: This is whats causing our tests to fail.

Its looking for a `src/` directory using `bridgetown/bridgetown-core` as the root. Should we perhaps point `site.source` towards the [./bridgetown-core/test/source](https://github.com/bridgetownrb/bridgetown/tree/main/bridgetown-core/test/source) directory to fix the failing tests? I dont think this code is itself the problem, but rather we need to adjust the tests.

ParamagicDev (changes requested):
Looks good. Super easy to follow, minimal changes made. If we could fix the failing tests (they actually produce errors not fail) we should be good to go.

It seems like we just have to adjust the value of `site.source` to fix where it searches for the `src` directory.

bridgetown-core/lib/bridgetown-core/frontmatter_defaults.rb
ParamagicDev: I cant think of a reason either, if youre descending, it has to be longer.

bridgetown-core/lib/bridgetown-core/readers/defaults_reader.rb
jaredcwhite: Thanks for the pointer. I'm curious why this would cause the tests to break suddenly, but it should be a simple fix either way.

Issue #145, opened 2020-09-22
Add paginator to Page object and templates


jaredcwhite:
It feels weird that the paginator in Liquid templates is simply `paginator` but in the actual Ruby code it's `pager`. (And you have to use `page.pager` to access it in a Ruby template.)

Let's add `alias_method :paginator, :pager` to `Bridgetown::Page` and also add the paginator (if present) directly within Ruby templates (ERB, etc.) so it's closer to how Liquid has worked historically.

jaredcwhite:
Fixed in #148

PR #148, opened 2020-09-23
feat: alias pager to paginator

ParamagicDev:

## Summary

Aliases Pager to Paginator...wasnt sure where to get started adding tests for this. I couldnt find anything in the bridgetown-core test suite related to `pager`. I would love to expand more on this PR but need maybe a little more guidance.

## Context

Addresses #145

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/page.rb
+++ b/bridgetown-core/lib/bridgetown-core/page.rb
@@ -14,6 +14,7 @@ module Bridgetown
     attr_accessor :data, :content, :output
 
     alias_method :extname, :ext
+    alias_method :paginator, :pager
 
     # A set of extensions that are considered HTML or HTML-like so we
     # should not alter them,  this includes .xhtml through XHTM5.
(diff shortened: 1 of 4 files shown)


jaredcwhite:
I was about to say take a look at the `bridgetown-paginate` gem in the monorepo, but…actually, there are no tests there. Ooopsie! So I guess this is fine for now. I would also like to get the `paginator` variable added to `RubyTemplateView` along with `site`, `page`, etc.

ParamagicDev:
Okay now that I can work with 👍

ParamagicDev:
@jaredcwhite Rather than messing around with core, I decided to just go into the paginate gem and alias the `newpage.pager = Paginator.new` to `newpage.pager = newpage.paginator = Paginator.new` Im not really sure where to test this out other than to spin up a new site and try it out and that just feels...wrong

jaredcwhite:
Now I'm waffling on this entirely…the `pager` term is a holdover from Jekyll and nothing in the codebase outside the pagination gem directly uses `pager` to the best of my knowledge. Wondering if we should just rename to `paginator` outright (but that could potentially break a few sites out in the wild…)

ParamagicDev:
Perhaps we mark it deprecated and provide a warning in this release, and then in the following release remove it? @jaredcwhite

jaredcwhite:
That's for sure the way to go, @ParamagicDev. So let's make `paginator` the real name and `pager` can be the alias, and we'll remove in a future release. 👍

PR #155, opened 2020-10-01
build: add matrix testing for Ruby

andrewmcodes:

## Summary

I hit an issue on Ruby 3 (that I will open in a sec), and was hoping that GitHub may provide access to use it but I see not. I am going to try a few other things but I think some matrix testing would help make sure that people on lower ruby versions aren't affected in changes.

## Context

1min

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,10 +14,11 @@ jobs:
     strategy:
       matrix:
         ruby_version: [2.5.x, 2.6.x, 2.7.x, head, debug]
+    continue-on-error: ${{ endsWith(matrix.ruby, 'head') || matrix.ruby == 'debug' }}
     steps:
       - uses: actions/checkout@master
       - name: Setup Ruby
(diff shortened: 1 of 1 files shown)


jaredcwhite:
@andrewmcodes Thanks, this is a great idea! Especially at some future point where taking advantage of Ruby 2.7, 3.0, and beyond is a big priority but we need to make sure older installations won't break.

andrewmcodes:
Agreed. I'd be happy to remove the head stuff for now if you'd just like 2.5, 2.6, and 2.7 and ignore 3 for right now but winter is approaching 🚨

Maybe only test against head when merged into master? Then it's easy to track but won't confuse contributors

jaredcwhite:
@andrewmcodes I took out head so we'll just test Ruby 2.x versions for now. Clearly some additional work is required to support Ruby 3 (and I should open an issue for that), so once that's ready to roll we can add that into the matrix. Thanks again for putting this together!

PR #157, opened 2020-10-01
Configurable template engines on a per-site or per-document basis (WIP)

### LINKED ISSUE #132: feat: Ability to switch "default" template processors
I'm increasingly concerned about the future of Liquid as a robust open source project, and that paranoia combined with the recent work to add Tilt support (ERB, etc.) is causing me to rethink Bridgetown's long-term template strategy.

To that end, I think it would be worthwhile to add the ability to switch the default template processor. In other words, standard files that send in `.html` or `.md` or `.json` could in theory be processed through ERB or Slim (or something else super-secret I may or may not be working on 😉) without having to use those exact template extensions.

Alternatively, we bite the bullet and adopt the multi-extension approach like other Ruby frameworks…`blog-post.md.erb`, `projects.html.slim`, etc. Maybe that's better in the long run, but it's definitely a major backwards-incompatible change so we should consider it carefully.

jaredcwhite:
Fixes #132 

See ongoing discussion there for context.

## TODO:

- [x] Move Liquid processing to its own `Converter` object
- [x] Ensure `.liquid` files are processed with Liquid even when ERB/etc. is the configured template engine
- [x] Tests
- [x] Documentation

## Notes on Breaking Changes

We lost a couple of features as part of these changes. One of which was undocumented, the other frankly dangerous and incompatible with a plurality of template engines. These are:

1. Previously it was possible in Liquid for a child layout to set a front matter variable and a parent layout to access the child layout's variable value, aka `{{ layout.variable_from_child_layout }}`. That's no longer the case now…each layout has access to only its own front matter data. I couldn't think of a good use case for that to warrant the added complexity, and ~~furthermore it was undocumented~~ (oops, see below). (If you go to Jekyll's Layout docs, for example, it makes no mention of this feature.)
2. A few pre-render hooks were provided access to Liquid's global payload hash. That meant they could alter the hash and thus the data being fed to Liquid templates. The problem is that there was no visibility into those changes from any other part of the system. Plugins accessing actual page/layout/site/etc. data wouldn't pick up those changes, nor would other template engines like ERB. I think if a hook needs to alter data, it needs to alter actual Ruby model data, and Liquid's payload should *always* reflect that model data. So that previous feature has been dropped as well.

One *new* feature/fix for Ruby-based templates (ERB, etc.) is they can now access `layout` in page templates, not just in layout templates themselves. Brings parity with the Liquid behavior.


andrewmcodes:
@jaredcwhite I vote no on this for really only one reason. 

If I am a developer with a blog on Jekyll, and I am looking at switching to Bridgetown, this instantly creates the appearance that the syntax is ERB aka I am no longer interested in that time investment. 

To me, ERB, HAML, & Slim are tools for power users, and because of that, they should be behind flags, not the default. 

To me the real question is, if this feature drives away any folks from trying Bridgetown, is it worth the upside? I know that a little more reading will reveal they can switch, but to me the happy path should be the one of least resistance. I love that I can now use ERB (HAML moreso though 😄 ), but I don't think it should the default. If it was changed to be the default, I would vote a bump to 0.2 to be very clear this potentially break things (like if some moron created a script to create a site to test something bc he was too lazy to test it correctly, you know...in theory).

**TL;DR I am worried this will dissuade Jekyll users from changing to Bridgetown and should not be the default. If it does become the default, Bridgetown should be upgraded to `0.2'.**

jaredcwhite:
All good points… just to clarify, there's no plan to change the _default_ setting from Liquid — i.e., you can still `bridgetown new` and use Liquid everywhere just as before. The problem has been if you wanted to change a setting to "decouple" Liquid from the site entirely and use something else across the board, that wasn't really possible. This makes that possible as an option, not a new default.

I guess the concern could be that if somebody even sees an example (a tutorial or demo repo or whatever) of Bridgetown configured using something other than Liquid, that's maybe a turnoff, but that seems a little far-fetched perhaps.

jaredcwhite:
Regarding the (1) breaking change above, I actually just found a place where the previous behavior was documented, waaaayyyy down at the bottom: https://www.bridgetownrb.com/docs/variables

So that'll need to be revised.

jaredcwhite:
Merged in, but documentation is still required.

PR #158, opened 2020-10-01
Use HashWithDotAccess::Hash for all data/config hashes

### LINKED ISSUE #144: feat: Dot access for data hashes (front matter, YAML/JSON files, etc.)
I've been experimenting with ways to provide dot access (aka `page.data.title`, `site.metadata.twitter.handle`, etc.) within Ruby templates/plugins. By far I find it the most annoying aspect of transitioning from Liquid to ERB/etc.

Since Bridgetown already switched from plain hashes to using `ActiveSupport::HashWithIndifferentAccess`, the next logical step would be for that to provide dot access as well. Oddly, I wasn't able to find any widely-referenced example for how to add that, so this is my best attempt at a monkey-patch:

```ruby
ActiveSupport::HashWithIndifferentAccess.class_eval do
  def respond_to_missing?(key, *)
    return true if "#{key}".end_with?("=")

    key?[key]
  end

  def method_missing(key, *args)
    if "#{key}".end_with?("=")
       self["#{key}".chop] = args.first
    elsif self.key?(key)
      self[key]
    else
      default
    end
  end
end
```

My feeling is that this should be added as a feature out-of-the-box, but if there's some reason it doesn't sit right with anyone (language purity? performance?), the monkey-patch could be disabled with an ENV var or something.

There's also the issue of keys colliding with valid Ruby methods (couldn't have a front matter key called `each` for obvious reasons), but that's no different than with ActiveRecord and database column names, so I think most Ruby programmers will understand the caveats involved, and there's always the typical `[]` access method to fall back on.

jaredcwhite:
Closes #144 

```
site.data.authors.jared.website
=> "https://jaredwhite.com"
```

That's a thing of beauty!

- [x] Tests
- [x] Documentation

### DIFF
--- a/bridgetown-builder/lib/bridgetown-builder/document.rb
+++ b/bridgetown-builder/lib/bridgetown-builder/document.rb
@@ -8,7 +8,7 @@ module Bridgetown
       def initialize(site, path)
         @site = site
         @path = path
-        @data = ActiveSupport::HashWithIndifferentAccess.new
+        @data = HashWithDotAccess::Hash.new
       end
 
       def front_matter(data)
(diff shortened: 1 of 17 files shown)


jaredcwhite:
FYI this leverages a new gem I just published: https://github.com/bridgetownrb/hash_with_dot_access

Issue #140, opened 2020-09-10
feat: Additional Ruby template helpers


jaredcwhite:
Now that we have Ruby templates (ERB, etc.), it would be useful to investigate what new helpers we could add to the framework… Rails, Middleman, etc. all have lots of prior art to draw from. Here's but one example: https://middlemanapp.com/basics/helper-methods/

jaredcwhite:
We should look into any more helpers that need to be ported over from included Liquid tags as well…`class_map` for example.

jaredcwhite:
Some native Liquid filters are also "missing" in the Ruby context…reference:

https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb

jaredcwhite:
I think we're in good shape for now. Any future helpers can be tracked through new issues. Closing via #268

PR #164, opened 2020-10-20
Add link/url_for and link_to helpers

jaredcwhite:
Partially addresses #140 

- [x] Tests
- [x] Documentation


jaredcwhite:
We should make sure `link_to` accepts actual URLs, just like the Rails counterpart (aka `https://apple.com`, `/path/to/anything`).

jaredcwhite:
Works with Ruby objects as well as long as they respond to the `url` method.

PR #167, opened 2020-10-21
Obfuscate link filter

### LINKED ISSUE #166: feat: Email obfuscation
## Summary

Provide a liquid filter for obfuscation of email addresses (and possibly other types of contacts)

## Motivation

I'm creating a bridgetown site for a client and wanted to obfuscate email-addresses and phone numbers on the page

I took [react-obfuscate](https://github.com/coston/react-obfuscate/blob/master/src/obfuscate.js) as a reference and threw together the most basic liquid filter:

```ruby
class Filters < SiteBuilder
  def build
    liquid_filter "obfuscate_mail" do |input|
      output = "<a href=\"obfuscated\" "
      output += "style=\"unicode-bidi: bidi-override; direction: rtl\" "
      output += "onfocus=\"this.href = 'mailto:#{input}'\" "
      output += "onmouseover=\"this.href = 'mailto:#{input}'\">"
      output += "<script type=\"text/javascript\">document.write(\"#{input}\".split('').reverse().join('').replace('(', ')').replace(')', '('));</script></a>"
      output
    end
  end
end
```

If this addition is welcome, I can put up a PR in no time.

## Unresolved questions

- as of now, I only obfuscate email addresses. with a second parameter, the prefix (`mailto:`) should be easily configurable.

julianrubisch:

## Summary

This PR introduces a simple `obfuscate_link` filter to prevent bots from sending emails etc.

The implementation takes the simplest path possible, reversing the link text and using CSS reading-direction style attributes to display it correctly.

The link href is replaced by `obfuscated` and the original target is restored on mouseover or focus (touch friendly). 

## Context

Closes #166

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/filters.rb
+++ b/bridgetown-core/lib/bridgetown-core/filters.rb
@@ -120,6 +120,20 @@ module Bridgetown
       Addressable::URI.normalize_component(input)
     end
 
+    # Obfuscate an email, telephone number etc.
+    #
+    # @param input[String] the String containing the contact information (email, phone etc.)
+    # @param prefix[String] the URL scheme to prefix (default "mailto")
+    # @return [String] a link that is unreadable for bots but will be recovered on focus or mouseover
+    def obfuscate_link(input, prefix="mailto")
+      output = "<a href=\"obfuscated\" "
+      output += "style=\"unicode-bidi: bidi-override; direction: rtl\" "
+      output += "onfocus=\"this.href = '#{prefix}:#{input}'\" "
+      output += "onmouseover=\"this.href = '#{prefix}:#{input}'\">"
+      output += "<script type=\"text/javascript\">document.write(\"#{input}\".split('').reverse().join('').replace('(', ')').replace(')', '('));</script></a>"
+      output
+    end
+
     # Replace any whitespace in the input string with a single space
     #
     # input - The String on which to operate.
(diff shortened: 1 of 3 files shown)


jaredcwhite:
I like the idea @julianrubisch, but since it requires Javascript to work anyway, I'm thinking a more straightforward approach might be to use a cipher.

For example, this could encode the string from the Ruby side: https://esolangs.org/wiki/ROT13-Ruby

and then this could decode it back: http://jsfromhell.com/string/rot13

That's an approach recommended here anyway: https://perishablepress.com/best-method-for-email-obfuscation/

I think we should stay away from `document.write` as well…it doesn't play nicely with live DOM updates via Turbolinks, etc.

julianrubisch:
Understood. What would you use in place of document.write, since the implementation referenced in https://perishablepress.com/best-method-for-email-obfuscation/ uses it too?

jaredcwhite:
Maybe put a script tag immediately following the `a` tag and use `document.currentScript.previousSibling.textContent = "…"`?

julianrubisch:
Going forward with this, I realized I need the ROT47 algorithm to cover numbers and other characters. The decoding JS code is a bit more intricate, but I think it pays off!

julianrubisch:
Can I do anything to push this forward?

jaredcwhite:
@julianrubisch My bad, I didn't notice that you'd already committed the fix. 😄 Merging!

PR #174, opened 2020-10-29
Convert ERB and Beyond docs to ERB and use capture helper for liquid_render

jaredcwhite:
I thought it would be a good exercise to—y'know—make sure the "ERB and Beyond" documentation page is itself using ERB instead of Liquid. In doing so I discovered I needed to fix the `liquid_render` helper so it uses `capture`. Will require a point release for 0.18.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -59,9 +59,8 @@ Lint/Void:
   Exclude:
     - bridgetown-core/lib/bridgetown-core/concerns/site/configurable.rb
 Metrics/AbcSize:
-  Max: 21
+  Max: 24
   Exclude:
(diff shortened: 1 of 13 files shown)


jaredcwhite:
OK, I've had a love-hate relationship with RuboCop since this project started. As my first experience with both a large codebase and a complex linter setup predating my involvement, I really didn't want to rock the boat too much.

But `Metrics/AbcSize` is driving me freaking bonkers.

So I relaxed it a bit, which is already relaxed from the default RuboCop setting (17? SEVENTEEN??!!). I welcome feedback, but I'm very likely going to stick with this decision. :)

PR #188, opened 2020-11-21
Add inferred name attribute to default package.json

ayushn21:

✅ I've added tests (if it's a bug, feature or enhancement)
✅ The test suite passes locally (run `script/cibuild` to verify this)

## Summary

When we create a new Bridgetown site, the default `package.json` has a `name` attribute of `bridgetown-new-site`. 

I thought it would be a nicer developer experience if this name could be inferred from the path entered into the `new` command. So this change takes the last path component of the input path and sets that as the name in `package.json`.

### DIFF
--- a/bridgetown-core/test/test_new_command.rb
+++ b/bridgetown-core/test/test_new_command.rb
@@ -60,7 +60,7 @@ class TestNewCommand < BridgetownUnitTest
       static_template_files = dir_contents(site_template).reject do |f|
         File.extname(f) == ".erb"
       end
-      static_template_files << "/Gemfile"
+      static_template_files.push "/Gemfile", "/package.json"
 
       capture_output do
         Bridgetown::Commands::Base.start(argumentize(@args))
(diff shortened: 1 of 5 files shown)


bridgetown-core/test/test_new_command.rb
```diff
@@ -60,7 +60,7 @@ def site_template_source
       static_template_files = dir_contents(site_template).reject do |f|
         File.extname(f) == ".erb"
       end
-      static_template_files << "/Gemfile"
+      static_template_files.push "/Gemfile", "/package.json"
```
andrewmcodes, line 63: Definitely prefer your solution but fun fact:

```rb
static_template_files = []
=> []
static_template_files << "/Gemfile" << "/package.json"
=> ["/Gemfile", "/package.json"]
```

andrewmcodes (approved):
👏 I've been meaning to do this! Love it!

bridgetown-core/test/test_new_command.rb
ayushn21: Ah nice, cheers! Didn't know that was valid syntax :)

jaredcwhite:
Thanks @ayushn21! This is a great enhancement. :+1:

PR #191, opened 2020-11-26
The Great Unification (removing all div-ision-s)

jaredcwhite:
This is a project to replace virtually all div and span tags on the bridgetownrb.com website with more semantically-meaningful custom elements, or in a few cases other HTML5 elements. In addition it adds [linthtml](https://github.com/linthtml/linthtml) as a tool to help enforce usage of tags other than div/span in the future.

Why? Because we can! 😄 Also custom elements are awesome, and "div" or "span" mean absolutely nothing so boo.

### DIFF
--- a/bridgetown-website/.linthtmlignore
+++ b/bridgetown-website/.linthtmlignore
@@ -1,0 +1,1 @@
+src/_docs/components.md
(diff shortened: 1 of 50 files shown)


ParamagicDev:
Missed opportunity. This should've been called "The Great Unification"

You know, because we're removing all the div[ision]s.

jaredcwhite:
You got it!

PR #192, opened 2020-11-30
Add `mailto:`-`<a>`-tag to generated footer

pascalwengerter:

Interestingly, I've run the test suite locally and this change didn't break any (so the footer might not be 100% tested?)

## Summary

One of the first changes I've made to the project I'm setting up right now, perhaps interesting for other folks as well?! Properly linking the site email in the generated `<footer>`.

## Context

No correlated issue found.


jaredcwhite:
Yeah we don't really test the contents of the templates themselves when a site's first generated, as they're considered mainly throw-away. Maybe if we were to build more of a "default theme" out, it'd make sense to test that. Anyway, thanks for the contribution! LGTM

PR #202, opened 2020-12-09
Add Ruby 3.0 preview 2 to the build action

jaredcwhite:
Adding this to our test matrix, and if there are Ruby 3-specific issues with the test suite we can fix those and merge into this branch.


jaredcwhite:
OK, it looks like the only real issue here is a problem with processing Ruby Front Matter, which I've tracked down to an issue with SafeYAML and our `Rb` string subclass. To be investigated…

PR #206, opened 2020-12-11
Add modules resolve paths to default webpack config

ayushn21:

## Summary

This PR adds 3 paths to Webpack's default module resolution list. This means that JS modules located in the following directories can be `import`ed or `require`d without specifying the full or relative paths.

* `frontend/javascript`
* `frontend/styles`
* `src/_components`

I also changed the documentation on the **Components** page to indicate that JS modules located there can be imported without specifying the path.

I know the current description describes a configuration where all JS files in the `_components` directory would be autoloaded.  But since those files can now be `import`ed with 0 config, I thought changing it was the more user-friendly approach. I'm happy to revert/rework the docs as per feedback though. Just did what I personally thought was appropriate!

## Context

https://github.com/bridgetownrb/bridgetown/issues/29#issuecomment-732454477

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby_version: [2.5.8, 2.6.6, 2.7.2]
+        ruby_version: [2.5.8, 2.6.6, 2.7.2, 3.0.0-preview2]
     continue-on-error: ${{ endsWith(matrix.ruby, 'head') || matrix.ruby == 'debug' }}
     steps:
       - uses: actions/checkout@master
--- a/bridgetown-core/lib/bridgetown-core.rb
+++ b/bridgetown-core/lib/bridgetown-core.rb
@@ -58,6 +58,16 @@ SafeYAML::OPTIONS[:suppress_warnings] = true
 class Rb < String; end
 SafeYAML::OPTIONS[:whitelisted_tags] = ["!ruby/string:Rb"]
 
+if RUBY_VERSION.start_with?("3.0")
+  # workaround for Ruby 3 preview 2, maybe can remove later
+  # rubocop:disable Style/GlobalVars
+  old_verbose = $VERBOSE
+  $VERBOSE = nil
+  SafeYAML::SafeToRubyVisitor.const_set(:INITIALIZE_ARITY, 2)
+  $verbose = old_verbose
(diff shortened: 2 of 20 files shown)


jaredcwhite:
I feel like the "sidecar JS assets" for components should just automatically load, since ideally every component's JS is independent thus the load order is irrelevant. Sort of like how Stimulus JS controllers all autoload as well.

Other than that, I approve these changes. :+1:

ayushn21:
Hmmm, yeah I agree that experience would be better. I'll update the PR soon :)

ayushn21:
The way I see it, if we want to autoload all `.js` files in `_components`, adding that directory to the search paths doesn't make a lot of sense as we'll never be manually `import`ing files from there.

In my last commit, I just took the approach currently on the Bridgetown website to import all component `js` files and added it to the default template so the developer doesn't have to do anything.

So the PR adds `frontend/styles` and `frontend/javascript` to the module search paths so the developer can `import` modules from those directories without specifying the path. It also adds a couple of lines to `frontend/javascript/index.js` to autoload all `js` files from `src/_components`.

What do you feel about this approach @jaredcwhite?

jaredcwhite:
This looks great @ayushn21! I'd forgotten about that example code to add the `liquidComponents` alias…and actually we're in the process of transitioning away from _only_ Liquid components to a polyglot configuration so that should probably be renamed to something else. `components` is likely too generic by itself. `srcComponents`? `localComponents`? Hmm…

ayushn21:
Hmmm tricky one. How about just `components`? Or `bridgetownComponents`?

jaredcwhite:
`bridgetownComponents` sounds good to me…no obvious collision with any other package names that might surface.

ayushn21:
Cool! Was thinking a bit more and also thought of `viewComponents` but that I guess would clash with https://github.com/github/view_component.

How about `bridgetownViewComponents` or do you think `bridgetownComponents` is better?

jaredcwhite:
> How about `bridgetownViewComponents` or do you think `bridgetownComponents` is better?

That is a little wordy…I think either `viewComponents` or `bridgetownComponents` is good, leaning towards the latter.

ayushn21:
Cool; I like `bridgetownComponents` as well so will make that update

ayushn21:
@jaredcwhite I've renamed the components alias. I reckon it's pretty much good to go :)

jaredcwhite:
LGTM! :+1:

PR #209, opened 2020-12-13
Create fedora.md

bkmgit:
Add instructions for Fedora.

Node and Yarn are in the Fedora repositories, so put instructions here, rather than using {% render "docs/install/node_on_linux" %} which is Ubuntu specific.

  - [x] I read the Project Goals and Future Roadmap pages at: https://bridgetownrb.com/docs/philosophy/
  - [x] I read the Code of Conduct: https://github.com/bridgetownrb/bridgetown/blob/master/CODE_OF_CONDUCT.md

## Summary

Add setup instructions for Fedora

## Context

Not related to any currently open issue

### DIFF
--- a/bridgetown-website/src/_docs/installation/fedora.md
+++ b/bridgetown-website/src/_docs/installation/fedora.md
@@ -8,6 +8,46 @@ order: 0
 
 ## Install Ruby
 
+### Using Rbenv
+
+Update your package list:
+
+Then install dependencies:
+
+```sh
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Thanks @bkmgit for the PR! I would like to request one change: can you test and document installing Ruby via rbenv in Fedora instead of (or alongside) the builtin Ruby approach? We generally don't recommend tieing the Ruby installation to whatever package is provided by the OS distribution.

bkmgit:
Thanks for the suggestion. Will make the changes and update the 
documentation later this week.

jaredcwhite:
Thanks for working on that @bkmgit. LGTM! 👍

Issue #207, opened 2020-12-12
docs: Add information about how to customize PostCSS


jaredcwhite:
With the PostCSS config that's landed in main, I'd like to reference useful documentation on how to configure it to support various features…nesting, custom media query variables, environment variables, etc. It looks like preset-env provides a lot of functionality out of the box, but it probably requires changing the "stage: 3" config to something else or adding an allowlist of desired features. There may be additional plugins out there that are popular as well we could link to.

ref: http://preset-env.cssdb.org/
https://github.com/csstools/postcss-preset-env#features
https://github.com/csstools/postcss-preset-env/blob/master/src/lib/plugins-by-id.js#L36

andrewmcodes:
I agree this could be helpful but think we need to draw a line at regurgitating the already very thorough postcss docs. 

My proposal would be to create “recipes” instead of arbitrary link. A recipe would be a description of how to accomplish an end goal vs blindly including things. 

Ex:
—-

# CSS Minification via cssnano 

## Prerequisites:
- PostCSS is installed 
- postcss-loader is installed and configured for the env 

## Directions: 

Make the following change to your PostCSS Config file: 

```diff
+ const cssnano = require("cssnano")({
+    preset: "default"
+ });

module.exports = {
    plugins: [
        // ...
        // This should be at the end of your plugin list
+        ...(process.env.NODE_ENV === "production" ? [cssnano] : [])
    ]
};

## Additional Resources

- https://cssnano.co/
```

andrewmcodes:
Sorry for formatting, I’m on my phone. 

But doing something in this format would answer `How?` vs begging the user to ask `Why?`

- why would I need postcss import 
- why would I need to do x y z

Telling them how to accomplish a task will reduce decision fatigue, be easier to search for, and could give us a nice template for other things that could be recipes like `How to add Tailwing`. 

Thoughts? 

I will re look at this on a computer and fix any typos or weirdness that occurred btw 😬

jaredcwhite:
I'm tracking with everything you said and I love your recipie example for cssnano. Not sure how to slot it into the current docs structure, but that's merely an implementation detail. Let's do it!

ayushn21:
Just thinking out loud; how about we bundle in automations for some common PostCSS configurations into Bridgetown? So instead of documenting "how" a user could do something and describing the config; we could just say "run this command".

[...]

PR #217, opened 2020-12-22
Add configure command to bridgetown to run packaged automations

ayushn21:

## Summary

This PR adds a `configure` command to `bridgetown`. The purpose of this would be to allow us to package automations with bridgetown itself so common tasks such as installing TailwindCSS or setting up other PostCSS plugins or combinations thereof can be done easily. 

I've added a `configurations` folder where all packaged automations can go. The `configure` command will just look in this directory for a file that matches the argument passed in; and if one exists, it will invoke the `apply` command passing that file as an argument.

Configurations can also be applied with `bridgetown new` using the `-c` flag. 

## TODO

- [x] Clean up code and improve syntax if/where possible
- [x] Compile a list of automations we want to package
- [x] Implement and add those automations
- [x] Add tests
- [x] Add docs

## Context

https://github.com/bridgetownrb/bridgetown/issues/207#issuecomment-749609164


jaredcwhite:
Holy guacamole, that was fast. This looks very cool @ayushn21 …as far as naming goes, addin? addon? to-may-to? tom-ah-to? We can workshop it a bit, but the idea is sound.

jaredcwhite:
Could also somehow tie into `bridgetown new` so you could run N additional installations after the site's created, just like the `--apply` option

ayushn21:
Cheers @jaredcwhite! Yeah I'll have a think about the naming over the next few days. Figured since I broke my resolution to stay away from code until 2021 today; I might as well make the most of it!! 

> Could also somehow tie into bridgetown new so you could run N additional installations after the site's created, just like the --apply option

Yeah for sure. Good idea!

DamnedScholar:
"Automata" would be a fun name, those being the actors that perform automations.

ayushn21:
Thanks @DamnedScholar, that's an interesting word. Hadn't heard it before!

However, thinking about this is bit more I'm inclined to rename the command `bridgetown configure`. Setting up stuff like Tailwind or an omakase postcss setup etc feel more like things we're configuring rather than things we're installing. `bridgetown configure tailwind` sounds more appropriate in my opinion.

So if there's no objections I think I'll rename the command to `configure` and the entities we're configuring can just be called `configurations`.

ayushn21:
I think I'm pretty much there with the implementation of the `configure` command. I think the next step for this PR should be adding in a few common configurations. `tailwind` and `netlify` I reckon are no-brainers. I'd also like to include an omakase PostCSS setup, though this might be a bit more tricky to get some kind of consensus on!

@jaredcwhite What do you reckon is the best next step with regard to the configurations we want to include?

I also need to update the docs. I think the best way would be to create new page in the `Configuration` section right after "Automations". 

Screenshot 2021-01-03 at 10 36 49 pm

ayushn21:
I reckon this one's good to go now. I've added in all the automations from [bridgetownrb/automations](http://github.com/bridgetownrb/automations) so I guess we can deprecate the repo once this is released.

I've also added in a `tailwindcss` configuration.

I still believe we absolutely must have a Bridgetown recommended PostCSS plugin setup. I think we should do that in a separate PR though as it'll be tricky to find consensus on the plugins that should be included.

I've added a new docs page called "Bundled Configurations" that describes this feature.

@jaredcwhite let me know what you think and if you'd like any changes :)

jaredcwhite:
Quick comment: I think we add should add Turbo and Stimulus (separately) as configurations. Arguably more important than Swup, so seems logical. Stimulus in particular would need a bit of a JS harness to set up… @ParamagicDev put together an automation a while back that could prove useful here. https://github.com/ParamagicDev/bridgetown-automation-stimulus/blob/master/bridgetown.automation.rb 

Let me know if you'd like any assistance.

ayushn21:
Yup I agree. Only added in Swup because it was already in the bridgetown/automations repo :) 

I'll add in both Stimulus and Turbo as well 👍.

jaredcwhite:
Code-wise this all looks great to me. Feel free to merge in any time. :+1:

PR #225, opened 2021-01-16
Change the default branch name of new bridgetown sites and plugins to `main`

ayushn21:
Inspired by the Rails team changing their repo's default branch to `main`; I thought we should make it so all new Bridgetown sites and plugins have a `main` branch by default.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/new.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/new.rb
@@ -158,7 +158,7 @@ module Bridgetown
         unless Bridgetown.environment == "test"
           inside(path) do
             run "git init", abort_on_failure: true
-            run "git checkout -b main"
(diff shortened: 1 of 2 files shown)


jaredcwhite:
I like the sentiment here. Git does have a way to configure the default branch name at the environment level, so for example on my computer I'll get main now any time I `git init`. I'm a wee bit wary of circumventing that…

ayushn21:
Fair point @jaredcwhite, I hadn't thought of that. I'm far too lazy to change any default settings myself :P

We could potentially check if the default branch is called master and rename it to main if so. This line seems to work:

`run "if [[ -n $(git status | grep master) ]]; then git checkout -b main; fi"`

Do you think that's over-engineering and not worth it? Happy to close this PR if that line's likely to cause more issues than it solves.

jaredcwhite:
I think that's a reasonable compromise. Let's try it 👍￼

PR #226, opened 2021-01-19
Add Bridgetown recommended postcss plugins configuration

ayushn21:

## Summary

I've added a configuration for PostCSS plugins as discussed in https://github.com/bridgetownrb/bridgetown/issues/207.

I decided to go with this color adjustment plugin: https://github.com/postcss/postcss-color-function, instead of the one @jaredcwhite recommended. That one was logging a deprecation/incompatibility message with PostCSS 8. The one I included is also an official PostCSS plugin so in my mind that's better.

Any and all feedback appreciated :) 

## Context

Closes https://github.com/bridgetownrb/bridgetown/issues/207

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/bt-postcss.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/bt-postcss.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# rubocop:disable all
+
+TEMPLATE_PATH = File.expand_path("./bt-postcss", __dir__)
+
+begin
+  find_in_source_paths("postcss.config.js")
+rescue Thor::Error
+  error_message = "#{"postcss.config.js".bold} not found. Please configure postcss in your project."
+
+  @logger.error "\nError:".red, "🚨 #{error_message}"
+  @logger.info "\nFor new projects, you can use #{"bridgetown new my_project --use-postcss".bold.blue}\n"
+
+  return
+end
+
+plugins = %w(postcss-easy-import postcss-mixins postcss-color-function cssnano)
+
+say "Adding the following PostCSS plugins: #{plugins.join(' | ')}", :green
+run "yarn add -D #{plugins.join(' ')}"
+
+remove_file "postcss.config.js"
+copy_file "#{TEMPLATE_PATH}/postcss.config.js", "postcss.config.js"
+
+# rubocop:enable all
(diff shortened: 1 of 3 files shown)


jaredcwhite:
This looks good. I'll have to look into the alternate color plugin you mentioned…wondering now why I'd picked an older one.

BTW, I'd like to test the `stage` configuration a bit more. For some reason I had to choose `stage: 4` on a recent project using Bridgetown, but I can't recall why, so I'll need to fiddle with that and get back to you. Ideally my project would be able to use this stock config verbatim.

ayushn21:
> BTW, I'd like to test the stage configuration a bit more.

Sure; all it does is polyfill up and coming CSS standards. Stage 4 would only allow features that are already part of the web. I think the default setting actually is stage 2. More details here if you haven't seen this page already: https://cssdb.org/#staging-process

Take your time; and let me know if you run into any issues! The PostCSS setup on my portfolio site is almost the same as this one so all being well it "should" work fine! 🤞

jaredcwhite:
@ayushn21 OK, I know what it was now. =) At stage 2, or even 3, it's baking the initial CSS variable values into the stylesheet so everything's defined twice. As an example:

```css
font-family: Arial, sans-serif;
font-family: var(--sans-serif-fonts);
```

I didn't want that (don't need compatibility with IE) so I bumped up the stage. Maybe there's another way around it. Anyway, compatibility is perhaps something other people *will* want, so I'm not adverse to having stage 2 the default (which PostCSS says is their default actually), but maybe we can document this somehow.

ayushn21:
Ahhh yeah; I remember being completely flummoxed by that when I was building Chapter24 last year. I'm not quite sure it's worth documenting though because the plugin will be polyfilling a bunch of stuff. Custom properties are still a stage 3 feature (http://preset-env.cssdb.org/features#custom-properties) which is why they're being polyfilled unless you set the stage to 4.

Maybe we could adapt the current docs wording to say:

It will also configure [`postcss-preset-env`](http://preset-env.cssdb.org) to polyfill all features at [Stage 2 and above](http://preset-env.cssdb.org/features#stage-2); .............

That sentence makes it clear that all features displayed on the linked page will be polyfilled and hence there might be duplication of properties.

What do you think?

jaredcwhite:
That sounds good. I'd even add something like: "If you don't need certain polyfills for your use case, you can bump up stage to 3 or 4 (for example, custom properties won't get polyfilled if stage is set to 4)".

ayushn21:
Hey @jaredcwhite, did you want some more time to play around with this config or is this more or less good to go?

I just made the update to the docs to explain the polyfill stuff.

jaredcwhite:
Ship it! :shipit:

PR #229, opened 2021-01-27
Fix grammar error

marketerly:

## Summary

Edited the following sentence: 
"But you can write an API using any virtually any language or framework…Node, Python, Go, you name it."

To:
"But you can write an API using virtually any language or framework… Node, Python, Go, you name it."


ayushn21:
Nice catch! 👏

PR #235, opened 2021-02-05
Remove docs about turning off liquid processing in frontmatter

ayushn21:

https://github.com/bridgetownrb/bridgetown/pull/177 disallows usage of Liquid and ERB within the same file. As such I don't think the "Turning off Liquid processing" section (https://www.bridgetownrb.com/docs/erb-and-beyond#turning-off-liquid-processing) in the docs is valid any more.

If I've understood correctly; I think we should just be able to remove this section. Please correct me if I'm wrong @jaredcwhite :)


jaredcwhite:
That's right, we don't have Liquid processing of ERB anymore…good call!￼

PR #252, opened 2021-02-24
 Update test matrix to Ruby 3.0.0

katafrakt:

## Summary

Since Ruby 3.0 is out already, no need to test against release candidate.


jaredcwhite:
Thanks! One more thing to cross off the todo list 👍👍

PR #253, opened 2021-02-24
Exclude current post from LSI-powered related posts

katafrakt:
This is a 🐛 bug fix.

## Summary

It seems that when using LSI to get related post, the first one is always current post (because, well, it _is_ the best match). This PR fixes it by manually removing current post from the result set. What's interesting is that Jekyll has exactly the same code as Bridgetown, but seems unaffected (i.e. I'm not getting current post as related, but on the other hand I use a bit old version of jekyll + classifier-reborn, maybe it's due to that).


jaredcwhite:
Thanks @katafrakt! Good to enforce that it will never include itself.

PR #262, opened 2021-03-19
update gh actions to use native ruby caching.

ParamagicDev:
This is a 🙋 feature or enhancement (for CI)

I was bored and looking at source code while this CI test suite runs and noticed we werent using caching native to ruby/setup-ruby!


jaredcwhite:
LGTM! :+1:

PR #267, opened 2021-03-28
update start.js script to work on win10

erikyuzwa:
This is a 🐛 bug fix.

## Summary

Brand new user of Bridgetown (yay!) and couldn't get the default site loading on `localhost` by following the getting started steps. Turns out I'm on Windows 10 and it doesn't know how to handle `sleep 8;` within my shell. 

I updated the command to allow the `sleep` plus running the next command. Now it works on my Windows10 and my Mac, but I have no other platform to test / verify.

## Context

Couldn't find a related issue, but could also be my search-fu not working.

### DIFF
--- a/bridgetown-website/start.js
+++ b/bridgetown-website/start.js
@@ -9,8 +9,8 @@ const port = 4001
 /////////////////
 concurrently([
   { command: "yarn webpack-dev", name: "Webpack", prefixColor: "yellow"},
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Thanks @erikyuzwa — while we don't officially "support" Windows natively, this is perfectly valid in a Unix-type context as well. Merging!

erikyuzwa:
thanks so much!

PR #276, opened 2021-04-04
(WIP) Fast Render mode

### LINKED ISSUE #123: feat: Fast Render mode
**September 2021 Update:** this feature has gone through a few conceptual permutations…not sure it's still going to be called this instead of just an "SSR" mode. More details forthcoming shortly.

----

**March 2021 Update:** this feature is now actively in the works and has been renamed to "Fast Render". The Resource content engine overhaul will aid in the development of this feature. The "global ID" idea below is now part of the resource origin system, soon to roll out in Bridgetown v0.20. Stay tuned for this to all come together Q2 2021!

----

## Summary

Instead of trying to improve upon the iffy-implementation of incremental rebuilds we inherited from our predecessor project, I propose scrapping that and focusing on a similar but really quite different use case: Fast Previews.

A "fast preview mode" would switch Bridgetown to an optimized rendering pipeline that's geared towards rendering only a single route. That single route could be based off of an actual content file in the repo, or it could be something programatic (aka from a headless CMS).

## Motivation

This feature would be useful for two contexts:

* Editing a page in a CMS-like environment and seeing a live update of the page quickly.
* Refreshing stale content served by a CDN with fresh content (for example, you could load a product page off a CDN and then immediately request the updated product page, perhaps using a serverless function).

## Guide-level explanation

You would use fast preview mode in one of two ways:

* Pass route/ID information (see below) as a command line argument to the `bridgetown build` command
* Load Bridgetown as a gem into your Ruby-based server (Sinatra, Rails, whatever) or serverless function and run a preview build command directly with the route/ID

The hardest part will be determining a 1:1 correlation between a final generated route and its source. Jekyll, and Bridgetown likewise, is completely uni-directional. In other words, it starts from source materials and that all ends up as destination material, but the destination material doesn't exactly "know" how it got there from the original source material.

I'm contemplating the idea of adding a new concept called a Global ID. (Yes, very similar in nature to [that feature in Rails](https://github.com/rails/globalid).) Every piece of page/document content would have a Global ID associated with it. That Global ID would get passed to the preview command and it would have logic to determine how that Global ID is related to either (a) content in the repo source, or (b) programmatic content that is handled by a plugin. Plugins that pull content would need logic to detect if Bridgetown is in fast preview mode, and if so, use the Global ID to figure out how to get just that content. For example:

Say there's a photo gallery which pulls data from the Strapi headless CMS using a plugin. On an individual photo page, it would have a Global ID of something along the lines of `gid://strapi/photos/152`. In the plugin, it would have some code like:

```ruby
if site.fast_preview? && site.previewing.global_id.start_with?("gid://strapi/photos/")
   # load the individual photo and generate the document
else
  # load all the photos in a normal build
end
```

How does a CMS editor or production site know what the Global ID is of a piece of content in order to request a refresh? We'd simply add the global id to the metadata of a page/document which could be inserted into the final HTML output where necessary. For example

```html
<body class="photo-page" data-global-id="gid://strapi/photos/152">
<!-- photo -->
</body>
```

## Drawbacks

Other than added complexity and probably a million edge cases I'm not thinking of, I don't see any major drawbacks to this approach. Normal/full builds would work exactly as they do now. Fast preview mode would simply alter the build pipeline so it knows it doesn't need to do a full build and can instead render out only what's requested — rendering out meaning either rewrite the output file in the destination folder or pass it back as a string (aka for use in a web API).

## Unresolved Questions

Eventually this could perhaps be combined with the dev server/watch process so when you are editing a file on disk and save it, that file would get updated in fast preview mode instead of going through a full build…starting to get back to the idea of incremental regeneration in that sense.

Another issue: in fast preview mode, none of the usual collections are likely to be loaded/processed, so if the page you're fast previewing is supposed to load a list of other collection items, that list would be blank. Perhaps in future there could be something done with caching to remedy that, but I don't feel it's a showstopper for getting a v1 out the door.

jaredcwhite:
Resolves #123 

Just like it says on the tin.

```ruby
> require "bridgetown"
> Bridgetown.fast_render("repo://products.collection/_products/books/back-from-the-future.md").content
=> "<h1>Back from the Future</h1><h2>A Celebration of the Greatest Time Travel Story Ever Told (Back to the Future Time Travel Facts and Trivia)</h2><p>A book by Brad Gilmore</p> …"
```

Still some performance/pipeline/exception issues to iron out, but this is very exciting. Want to use Bridgetown for your Rails view layer?

```ruby
class ProductsController < ApplicationController
  def show
    id = "repo://products.collection/_products/books/#{params[:id]}.md"
    render html: Bridgetown.fast_render(id).content
  end
end
```

(not quite that simple in practice…need to handle 404s gracefully, probably should have a dedicated Origin ID builder that can accept parameters, etc., etc.…but you get the idea)

**Note:** this PR changes `FileOrigin` (`file://`) to `RepoOrigin` (`repo://`). Thought that is both clearer and doesn't compete with the actual `file` URL scheme.


jaredcwhite:
BTW, in case you're not familiar with the new origin system in the Resource content engine, it means we can reference content coming from any number of possible sources. For example, if you were to use Strapi as a headless CMS, you could set up a Strapi origin and then you'd fast render like so:

```ruby
Bridgetown.fast_render("strapi://products/back-from-the-future").content
```

under the hood, the Strapi origin plugin would do a GET request to the API and render the proper template:
```
https://strapi-server.anywhere/products?slug=back-from-the-future
```

jaredcwhite:
I think I'm going to rework this a bit and file a new PR. Currently trying out some stuff within the context of a real-world site build, so whatever ends up working best there I'll extract fresh.

Issue #278, opened 2021-04-06
feat: Use the sass package versus node-sass


eclectic-coding:
## Summary
So, Libsass, which `node-sass` is a binding of, has been [deprecated](https://sass-lang.com/blog/libsass-is-deprecated) since October 20, 2020, with no new features having been added since November 2018. Therefore, according to the [Sass-lang](https://sass-lang.com/dart-sass) website:
> Dart Sass is the primary implementation of Sass, which means it gets new features before any other implementation

We might want to consider switching the default installation of Bridgetown to use `sass`. 
#273 
## Implementing
I have implemented on my Bridgetown 0.20 project and there would been to be a few tweaks to webpack, but other than that it is pretty straight forward and everything works perfectly. 

Remove `node-sass` and add these new packages as dev dependacies:
```
"fibers": "^5.0.0"
"sass": "^1.32.8",
```
Change the `sassloader` configuration in webpacker as such:
```
{
    loader: "sass-loader",
    options: {
      implementation: require("sass"),
      sassOptions: {
        fiber: false,
        includePaths: [
          path.resolve(__dirname, "src/_components")
        ],
      },
    },
  },
```
If you decide this is a direction you want to proceed, I will be glad to open a PR.

jaredcwhite:
> I have implemented on my Bridgetown 0.20 project … I will be glad to open a PR.

Yes, please do! Thanks @eclectic-coding

jaredcwhite:
Fixed in #279

Issue #273, opened 2021-04-04
feat: Replace the deprecated node-sass with (dart)sass


andreimoment:
## Summary

node-sass has been deprecated. Please replace with the sass package - https://www.npmjs.com/package/sass

## Motivation

This would bring sass functionality  in sync with the latest / up to date version of sass

jaredcwhite:
Fixed in #279

PR #279, opened 2021-04-06
Migrate to dart sass since node-sass is deprecated

eclectic-coding:
This is a 🙋 feature or enhancement, addressed in issues #278 and #273 to update the deprecated `node-sass` package.

### DIFF
--- a/bridgetown-core/lib/site_template/package.json.erb
+++ b/bridgetown-core/lib/site_template/package.json.erb
@@ -27,10 +27,10 @@
     <% if options["use-postcss"] %>
     "postcss": "^8.1.9",
     "postcss-flexbugs-fixes": "^4.1.0",
-    "postcss-loader": "^5.0.0",
+    "postcss-loader": "^4.1.0",
     "postcss-preset-env": "^6.7.0",
     <% else %>
-    "fibers": "",
+    "fibers": "^5.0.0",
     "sass": "^1.32.8",
     "sass-loader": "^8.0.2",
     <% end %>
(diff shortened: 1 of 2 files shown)


bridgetown-core/lib/site_template/package.json.erb
```diff
@@ -27,11 +27,12 @@
     <% if options["use-postcss"] %>
     "postcss": "^8.1.9",
     "postcss-flexbugs-fixes": "^4.1.0",
-    "postcss-loader": "^4.1.0",
+    "postcss-loader": "^5.0.0",
```
jaredcwhite, line 30: I'd like to upgrade PostCSS but let's not include it in this PR as it'll require additional testing unrelated to Sass.

bridgetown-core/lib/site_template/package.json.erb
```diff
@@ -27,11 +27,12 @@
...
     "postcss-preset-env": "^6.7.0",
     <% else %>
-    "node-sass": "^4.13.1",
-    "sass-loader": "^8.0.2",    
+    "fibers": "",
```
jaredcwhite, line 33: Can we take this out if we're not using fibers by default?

jaredcwhite (changes requested):
Thanks @eclectic-coding! Just a couple of comments/questions and then we should be good to go.

bridgetown-core/lib/site_template/package.json.erb
eclectic-coding: Okay not sure how that happened but will push a fix.

bridgetown-core/lib/site_template/package.json.erb
eclectic-coding: Fibers is needed along with sass-loader

eclectic-coding:
All requests have been resolved

jaredcwhite:
I took fibers out of the package.json since it's optional and not used in this configuration.

eclectic-coding:
Fibers is indicated in the wbpack.config.
And is recommended for performance 
https://github.com/sass/dart-sass#javascript-api

Issue #269, opened 2021-03-31
feat: Switch to Rack + Roda and away from Webrick


jaredcwhite:
As part of the push towards ~~Rails API~~ Ruby backend integration, it's become clear to me we need to remove our dependency on Webrick and retool around Rack (and Puma as the actual server). And not just Rack alone but a routing layer which can sit just above it—a good solution being [Roda](http://roda.jeremyevans.net/index.html).

This would provide a number of benefits:

* The only use case currently for Webrick is local site development/testing. Webrick isn't recommended for any production use. While Bridgetown typically is a build-and-deploy solution (hence the Static Site Generator moniker), there are reasons why running Bridgetown "live" through a web service could be desirable.
* Furthermore, by making Bridgetown essentially a Rack app, it not only makes integration with an API easier, it allows Bridgetown itself to enter the realm of true all-in-one web framework…with the ability to handle a pretty seamless range of SSG & SSR needs.
* Alternatively, if you already have a Rack app/Rails/etc., you could "bolt" Bridgetown on as a sub-path, just like any number of other apps (like how Rails apps can incorporate Sinatra apps such as Sidekiq's admin UI).

My initial preference to start would simply be to swap Webrick out for Rack/Puma and otherwise keep all features and CLI options as much the same as possible. Then we can assess where to go from there.

Feedback most welcome!

ayushn21:
I'm 100% behind the idea of swapping out Webrick for Puma or similar. Though I don't quite understand why we'd want Bridgetown to be an all-in-one framework. 

I'm just wary of losing a degree of focus and simplicity by stretching ourselves too thin and attempting to solve too vast an array of problems with Bridgetown. The way I see it, keeping Bridgetown as a build-and-deploy solution keeps the focus of the project strong. 

Combining Bridgetown with a first-class integration with Rails (which has one sole purpose of being a rock solid web framework) would in my opinion fulfil any use case that running Bridgetown live thorough web service would.

Just my $0.02.

What do you think the advantages of taking Bridgetown into the realm of an all-in-one framework would be?

jaredcwhite:
> What do you think the advantages of taking Bridgetown into the realm of an all-in-one framework would be?

I'm probably getting ahead of myself here…the primary use case I'm thinking of is being able to do a live rerender of a particular URL, maybe pass some request params along. Not trying to compete directly with Sinatra/Rails/etc. But thanks for the reminder we shouldn't try to dilute the value prop of the technology.

[...]

PR #281, opened 2021-04-08
Support for Rake, Rack and Roda (and Puma)

### LINKED ISSUE #37: Weird shell history issue after using sync.js
In iTerm, sometimes when I ctrl-c to quit the dev services, and then I press the up arrow key to browse shell history, instead of seeing the actual commands I just get garbage characters: `^[OA`

Wonder if the `sync.js` script needs to output a newline or something upon termination to get the shell to work correctly.

### LINKED ISSUE #280: feat: use Bridgetown CLI directly instead of Yarn?
I put a question mark at the end there because I'm not entirely sold on the idea myself, but I've been musing on the idea of making the `bridgetown` command itself handle the common use cases we're currently relying on `yarn` to do. Think:

* `bridgetown start` (or `dev`)
* `bridgetown deploy`
* `bridgetown frontend:compile`
* etc.

I also did a personal hackathon a couple of days ago to try out Rake/Roda and in the processes essentially recreated Concurrently/Foreman/etc. in order to run multiple processes in parallel from a single Ruby script (Bridgetown watch, Webpack, and Browsersync). I like the idea of removing dependencies there and handling all the processes directly. But `yarn` has a lot of institutional awareness already and it's easy to tweak and add your own scripts in `package.json`. On the other hand…we're a Ruby tool, so why are using Yarn instead of, oh I dunno, Rake or [Toys](https://github.com/dazuma/toys) or something?

jaredcwhite:
**June update**: this is now top priority for me, next to getting all docs/features wrapped up for the official Resource engine switch.

----

I'll need to circle back to fully document this, because holy moly this was complicated. But in the end, worth it.

* We get full Rakefile support with very slick Thor-based automations integration
* We move away from Webrick and onto Rack + Puma — get ready for SPEED
* Roda is used for handling basic static file serving of the `output` folder, but it can easily be augmented to handle all kinds of dynamic server abilities including mounting a Rails (or any other Rack-based) app
* Ditch all the `bundle exec` nastiness and use a binstub — I also want to document doing `alias bt="bin/bridgetown"` because then it's a cinch
* We can stop using `yarn` to bootstrap everything. `bt start` > `yarn start`.  Moar Ruby =)

Probably will want to put together an automation to upgrade old Bridgetown sites (otherwise it's a bunch of manual editing of files like `package.json` and adding in new stuff). Should look into how it could intersect with #250 

Resolves #280, #269
Closes #37


jaredcwhite:
I broke everything. :(

jaredcwhite:
Yay, deploying now on both Netlify and Render. :+1:

jaredcwhite:
Just making a note here

jaredcwhite:
Closing in favor of the newer #383.

PR #285, opened 2021-04-16
End-to-end Ruby front matter, templates, and data files

### LINKED ISSUE #263: feat: Super-powerful new Ruby Front Matter syntax
Suggested by @drush!

Instead of parsing front matter as YAML and then having to do a funky syntax to support Ruby expressions, we could simply add a native Ruby hash syntax. That way:

```yaml
title: I'm a Page
categories:
 - category1
 - category2
```

could instead be expressed by:

```ruby
{
  title: "I'm a Page",
  categories: [1, 2].map { |num| "category#{num}" }
}
```

For the front matter delimiter, I propose the following since it works well with Markdown syntax highlighting (in VS code anyway):

```
~~~ruby
{
  key: "value"
}
~~~

Content here.
```

If we wanted to be more in sync with the style Eleventy uses for dynamic front matter, we could do this:

~~~
---ruby
{
  key: "value"
}
---

Content here.
~~~

But that doesn't provide syntax highlighting in Markdown.

As it so happens **I've already implemented this locally** 🤯 (with the first option), and it works extremely well! It even evaluates lambdas in the context of the resource object so you can do stuff like this:

```ruby
{
  layout: :post,
  title: "Blog post",
  also_title: ->{ data.title },
  permalink: ->{ "/sections/#{collection.label}/#{data.title.gsub(' ', '').downcase}.html" }
}
```

Whoa.

jaredcwhite:
What would it mean if you could write pure Ruby for:

* Front matter of any resource
* Template of any resource
* Data files — yes, literally `src/_data/lots_of_data.rb`

That's what this PR does.

Resolves #263 

- [x] Tests (might need a couple more though)
- [x] Documentation


jaredcwhite:
Oh, something I was experimenting with (Markdown "content parts") got into this PR, I need to strip that out and address separately.

PR #293, opened 2021-05-04
Make configure command independent of Apply command

ayushn21:

## Summary

Currently the `bridgetown configure` command invokes `bridgetown apply` under the hood. I've tweaked it so the configure command now directly calls Thor's `apply` method to execute the bundled configurations.

Going forward this will allow us to expose utility methods and instance variables (like `@site` potentially) just to our bundled configuration scripts and keep them isolated from what we expose to third party automations.


jaredcwhite:
Nice, sounds good.

PR #294, opened 2021-05-04
Consider the default branch from the git config while create a new site or plugin

ayushn21:

## Summary

To set the default branch name in new bridgetown sites and plugins, we're currently using a flaky if/then line of shell script that checks if the branch name is `master` and updates it to `main` if it is.

A PR just got merged into Rails where this is implement in a more robust way that checks the git config for a default branch name first and sets the default branch to main only if it doesn't exist:

https://github.com/rails/rails/pull/40254/files

Copying the code from there I've improved the logic we use to set the default branch.


jaredcwhite (approved):
Looks good to me! :+1:

PR #296, opened 2021-05-06
Fix warnings in plugin tests by checking if an ivar was defined

ayushn21:
This is a 🐛 bug fix. 

## Summary

One of my plugins' tests were logging the following lines:

`frontmatter_defaults.rb:16: warning: instance variable @glob_cache not initialized`
`localizable.rb:8: warning: instance variable @locale not initialized`

This PR fixes those by checking if those instance variables were `defined?` rather than if they exist.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/concerns/site/localizable.rb
+++ b/bridgetown-core/lib/bridgetown-core/concerns/site/localizable.rb
@@ -5,7 +5,7 @@ class Bridgetown::Site
     # Returns the current and/or default configured locale
     # @return String
     def locale
-      if @locale
+      if defined?(@locale)
         @locale
       else
         @locale = ENV.fetch("BRIDGETOWN_LOCALE", config[:default_locale]).to_sym
--- a/bridgetown-core/lib/bridgetown-core/concerns/site/localizable.rb
+++ b/bridgetown-core/lib/bridgetown-core/concerns/site/localizable.rb
@@ -6,10 +6,10 @@ class Bridgetown::Site
     # @return String
     def locale
       @locale ||= begin
-        ENV.fetch("BRIDGETOWN_LOCALE", config[:default_locale]).to_sym
+        locale = ENV.fetch("BRIDGETOWN_LOCALE", config[:default_locale]).to_sym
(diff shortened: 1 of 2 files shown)


ayushn21:
I'm not sure what's up with that test failure @jaredcwhite. I can't reproduce it locally and I've rerun the tests a couple of times already. Any ideas?

jaredcwhite:
Hmm, sounds like somehow Ruby's warning level was set to the most strict. I normally don't see any warnings about missing ivars in my Ruby projects.

ayushn21:
It was probably set to the default for new Bridgetown plugins. I'm not sure I even know how to change the warning level :P 

I reckon it's worth fixing though because checking an instance variable before it's been defined seems a bit iffy in my mind 🤷 

Not too fussed either way though.

bridgetown-core/lib/bridgetown-core/concerns/site/localizable.rb
```diff
@@ -5,7 +5,7 @@ module Localizable
     # Returns the current and/or default configured locale
     # @return String
     def locale
-      if @locale
+      if defined?(@locale)
```
jaredcwhite, line 8: let's just memoize this via a `@locale ||= begin` block

bridgetown-core/lib/bridgetown-core/frontmatter_defaults.rb
```diff
@@ -13,7 +13,7 @@ def initialize(site)
     end
 
     def reset
-      @glob_cache = {} if @glob_cache
+      @glob_cache = {} if defined?(@glob_cache)
```
jaredcwhite, line 16: With a second look at this, I don't think it needs the `if` statement at all actually.

jaredcwhite (changes requested):
Couple suggestions on alternate fixes…

bridgetown-core/lib/bridgetown-core/concerns/site/localizable.rb
ayushn21: Ah nice, I didn't know you could use `begin` / `end` in this way!

ayushn21:
Updated the PR based on your feedback @jaredcwhite :)

PR #303, opened 2021-05-12
Remove safe_yaml gem in favour of using Psych which is in the stdlib

### LINKED ISSUE #300: Drop safe_yaml dependency in favour of using Psych directly
[`safe_yaml`](https://github.com/dtao/safe_yaml) doesn't seem to be maintained any more. It's logging a deprecation notice in its use of `Psych.load`: https://github.com/dtao/safe_yaml/pull/98 

I wonder if we could use [`Psych.safe_load`](https://www.rubydoc.info/stdlib/psych/Psych.safe_load) directly and drop the `safe_yaml` dependency altogether? 

I don't know if they're a like-for-like replacement though.

ayushn21:
This is a 🐛 bug fix. 

## Summary

The `safe_yaml` gem doesn't seem like it's maintained any more. The `Psych` module which is part of the stdlib has a `safe_load` method so we can drop that external dependency and use the stdlib directly.

## Context

Resolves #300


jaredcwhite:
This looks good! One question: is there a test to double-check it's really safe…like it can't instantiate a non-allowlisted object?

ayushn21:
> is there a test to double-check it's really safe…like it can't instantiate a non-allowlisted object

I know it's safe because I ran into that error a number of times while working on this hahaha. But yeah definitely worth adding a test for this. Will write a couple of tests for `Bridgetown::YAMLParser`.

ayushn21:
Added some tests. Let me know what you think!

ParamagicDev:
@jaredcwhite heres some background on `safe_load` https://github.com/ruby/psych/pull/487

jaredcwhite:
@ayushn21 Awssome! LGTM

Issue #315, opened 2021-05-21
The bt-postcss config disables Tailwind config from working


jaredcwhite:
**Bridgetown Version**:
0.21.0.beta3

**To Reproduce**:
If you run the `tailwindcss` bundled configuration, and then run `bt-postcss`, it seems like it breaks Tailwind from importing in the main CSS file.

**Expected behavior**:
Either we want it not to interfere with Tailwind working, or it should detect the existing presents of Tailwind and not continue with the repo changes.

ayushn21:
I can't ever see these two configurations being used in conjunction to be honest. Though I agree that this is a problem. 

I propose that we add a warning when both these configurations are run saying "This will overwrite your `postcss.config.js`, do you want to continue?".

In the future we could probably look at doing something a bit more clever, but it's non-trivial as the order of the postcss plugins is very important and our users could have any number of plugins in their config file prior to running this configuration.

jaredcwhite:
Yeah, I know it's non-trival to sort this out…all I know is I tried setting up a new project using both and then was scratching my head wondering why the `@tailwind` directives in the index CSS were just getting output to the bundle as-is and Tailwind was broken. So maybe a warning would be helpful at the least.

PR #317, opened 2021-05-24
Add confirmation for overwriting postcss config in tailwindcss and bt-postcss bundled configurations

ayushn21:

The `tailwindcss` and `bt-postcss` bundled configurations overwrite the `postcss.config.js` and the only place this is documented is on the website.

This PR adds a confirmation that informs the user that the `postcss.config.js` will be overwritten and asks if they would like to proceed.

Any confusion with running these two configurations one after the other should hopefully be prevented with this warning.

One area we should improve in the future is perhaps silence this warning when the config is invoked as part of `bridgetown new` but for now I reckon this should be good enough.

Closes https://github.com/bridgetownrb/bridgetown/issues/315

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/bt-postcss.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/bt-postcss.rb
@@ -13,12 +13,14 @@ unless File.exist?("postcss.config.js")
   return
 end
 
+confirm = ask "This configuration will ovewrite your existing #{"postcss.config.js".bold.white}. Would you like to continue? [Yn]"
+return unless confirm.upcase == "Y"
+
 plugins = %w(postcss-easy-import postcss-mixins postcss-color-function cssnano)
 
 say "Adding the following PostCSS plugins: #{plugins.join(' | ')}", :green
 run "yarn add -D #{plugins.join(' ')}"
 
-remove_file "postcss.config.js"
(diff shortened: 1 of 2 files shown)


bridgetown-core/lib/bridgetown-core/configurations/bt-postcss.rb
```diff
@@ -13,12 +13,14 @@
   return
 end
 
+confirm = ask "This configuration will ovewrite your existing #{"postcss.config.js".bold.white}. Would you like to continue? [Yn]"
+return unless confirm.upcase == "Y"
```
jaredcwhite, line 17: good opportunity to use `casecmp?`?

bridgetown-core/lib/bridgetown-core/configurations/tailwindcss.rb
```diff
@@ -11,13 +13,17 @@
   return
 end
 
+confirm = ask "This configuration will ovewrite your existing #{"postcss.config.js".bold.white}. Would you like to continue? [Yn]"
+return unless confirm.upcase == "Y"
```
jaredcwhite, line 17: here too?

bridgetown-core/lib/bridgetown-core/configurations/bt-postcss.rb
ayushn21: Ah yes, good call ... I'll update

PR #319, opened 2021-05-25
Install required packages in webpack enable postcss tool

ayushn21:
When enabling postcss using `bridgetown webpack enable-postcss`, it only updates the webpack config and doesn't install the required packages. Adding a step to install the required postcss packages and uninstall the sass packages.

I also pulled in the default postcss config from the site template instead of creating an empty file


ayushn21:
I give up on the test suite 😭. I can promise it passes locally and it's a false negative :(

ayushn21:
OMG it's gone green :D

jaredcwhite:
Looks good to me! :+1:

Issue #327, opened 2021-05-29
Post-install message via `bridgetown new` has vanished


jaredcwhite:
**Bridgetown Version**:
0-21-beta branch

The "Success!" message just isn't there. You see bundle and yarn install messages and then you get back to the shell prompt. Weird. This is definitely a blocker prior to the release of 0.21.

jaredcwhite:
So this line in `Bridgetown::Commands::New#create_site` is where the logger gets broken. `logger.info` works before it, doesn't work after it.

```ruby
invoke(Webpack, ["setup"], {})
```

@ayushn21 I think you mentioned you were having an issue with testing and getting proper log output? Maybe it wasn't a test problem but an actual bug?!

ayushn21:
Ahhh, I could have sworn it was working in the terminal and failing only in the tests. I've just looked again and I wasn't looking closely enough. We're doing `bundle install` and `yarn install` after invoking `Webpack`

```
 @skipped_bundle = true # is set to false if bundle install worked
bundle_install path unless options["skip-bundle"]

@skipped_yarn = true
yarn_install path unless options["skip-yarn"]
```

So I just saw that stuff was logging out and didn't look closely enough to see our success messages weren't being logged. Sorry, my bad!

I just ran `bridgetown new` with the `-c` flag and that doesn't log out our success message either. It seems the logger breaks whenever we call `invoke`. No clue why that might be though.

I can take a peek at it tomorrow at some point.

ayushn21:
> I also just ran bridgetown new with the -c flag and that doesn't log out our success message either. It seems the logger breaks whenever we call invoke. No clue why that might be though.

Sorry, this isn't the case. I ran that test well past midnight last night so didn't remove the invocation of Webpack as brain as gone to sleep! 🤦 

Looks like invoking the webpack command is doing something funky to the logger; looking into it now

ayushn21:
This is the offending line of code in `webpack.rb`:

```      
def site
  @site ||= Bridgetown::Site.new(configuration_with_overrides(quiet: true))
end
```

Any idea why creating an instance of the site might cause the logger to break? I can work around it for now, but I reckon it'll be good to understand why this happens. Maybe it was also the cause of the suddenly flaky test suite?

jaredcwhite:
Fixed in #328, closing.

PR #328, opened 2021-05-30
Fix the Bridgetown logger and other test improvements

ayushn21:
I broke the Bridgetown logger by setting `quiet: true` in the Webpack command, not realising that it would globally disable logging. Sorry, my bad and I should have caught it earlier.

In this PR I've un-disabled the logging and made a number of other minor tweaks to the tests. The great news is that it seems the test suite was unstable because logging was being disabled so we can close that issue as well. 

I also reverted the `test_new_command` to an earlier state when it was checking the log output.

Resolves https://github.com/bridgetownrb/bridgetown/issues/327


ayushn21:
Hmmm, so it seems the test suite is stable again locally but not in GH actions .... I'll leave that issue open for now

jaredcwhite:
Yeah I don't know understand why it keeps crashing in the middle of the test suite. Anyway, thanks so much for digging into this again and I'm glad we figured out the "spooky action at a distance"!￼ 😅

bridgetown-core/lib/bridgetown-core/configuration.rb
```diff
@@ -226,7 +226,7 @@ def read_config_file(file)
         raise ArgumentError, "Configuration file: (INVALID) #{file}".yellow
       end
 
-      Bridgetown.logger.info "Configuration file:", file
+      Bridgetown.logger.debug "Configuration file:", file
```
jaredcwhite, line 229: So by changing this from info to debug, it means that by default the configuration file line is never output to the user's terminal (they'd have to pass the --verbose flag).￼ I'm not necessarily opposed to this, but it's an obvious change from previous behavior.

ayushn21: My thinking here was that it's not really super important information so I reckon it's ok to bump it down a level. The reason I did it now is because it gets logged out when creating a new site. This is because the `webpack` command creates a configuration object that reads the file. 

This was the reason I inadvertently turned off logging earlier; I was trying to get this one line to stop being logged out when `bridgetown new` is run.

Do you think it's something that we absolutely should be logging out without the verbose flag?

jaredcwhite: I've found it annoying in some of the work I've done with SSR prototypes as well. Let's do it and just make a mention in our changelog/release notes. :+1:

Issue #318, opened 2021-05-24
Improve stability of the test suite


ayushn21:
Recently, I've found I need to re-run our test suite several times in PRs to get it to go green. I'm not sure what's caused the recent jump in false failures but I think it's something we should sort out sooner rather than later.

I get a fair amount of false failures locally as well.

Unfortunately I'm not sure what's causing the false negatives or how to investigate and improve it. I thought I'd open an issue for it so we can track it.

jaredcwhite:
I've seen that as well. It seems like something related to an autoloader issue (aka it's not doing the proper cascade of local content overriding plugin content). I've never been able to reproduce it locally though, only in GH actions. :(

ayushn21:
It's not just that issue, there's quite a few random "crashes" happening as well ... there aren't any failing tests logged out, just says the command failed

```
+ ruby -S bundle exec rake TESTOPTS=--profile test
23
[...]

PR #336, opened 2021-06-11
Attempting to stabilize test suite

ayushn21:
Looking into the test failure, it looks like the suite crashes when trying to create a new site in `test_webpack_command` and `test_new_command` due to a directory conflict.

In some cases, the new site's directory isn't totally cleaned up when running the whole test suite; not the individual case. I have no idea why it doesn't get cleaned up, can't see anything in to code to suggest it. 

Locally, I changed the hardcoded name of `new-site` and replaced it with `SecureRandom.alphanumeric` and it seemed to pass consistently. I think this is the right way to go as it would stand us in good stead if we were to parallelize the test suite in the future.

If anyone has any ideas as to why the directory isn't cleaned up sometimes, please let me know!

Opening this PR to see if CI passes consistently.

Resolves https://github.com/bridgetownrb/bridgetown/issues/318 (Hopefully)

### DIFF
--- a/bridgetown-core/test/test_new_command.rb
+++ b/bridgetown-core/test/test_new_command.rb
@@ -31,143 +31,141 @@ class TestNewCommand < BridgetownUnitTest
     end
   end
 
-  unless ENV["GITHUB_ACTIONS"]
-    context "when args contains a path" do
-      setup do
-        @path = "new-site"
-        @args = "new #{@path}"
-        @full_path = File.expand_path(@path, Dir.pwd)
-        @full_path_source = File.expand_path("src", @full_path)
-      end
-
-      teardown do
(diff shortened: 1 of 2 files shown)


ayushn21:
Hmmm, `TestRegenerator#test_: The site metadata should store modification times. :` keeps failing 🤔

ayushn21:
`TestComponents#test_: basic Ruby components should allow source components to override plugin components. ` also seems like a slightly flaky test

ayushn21:
Running the test suite a number of times, it seems that crash we were seeing as stopped. The above two tests fail on occasion. It's definitely something we should address but I reckon it's rare enough to procrastinate ;)

jaredcwhite:
Awesome, fix looks great! LGTM

PR #340, opened 2021-06-18
Add webpack merge docs  to webpack config

ayushn21:
I've been knee deep in Webpacker 6 today and I discovered the amazing [`webpack-merge`](https://github.com/survivejs/webpack-merge) package to merge config objects.

I thought it'd be a good idea to include it in new sites by default and add some docs about how to use it.


jaredcwhite:
Sounds awesome!

PR #344, opened 2021-07-02
Resource extension API along with basic summary feature

### LINKED ISSUE #306: Add summarization (auto-excerpt) to resources
The legacy content engine has an auto-excerpt functionality. Recent conversations in chat raised the point that we should provide some kind of API for "summarization" of resources, and the exact algorithm used is then a third-party concern. (For example, the ["separator"-based method](https://www.bridgetownrb.com/docs/posts#post-excerpts) of excerpting is something I never use personally. I actively dislike it, in fact. I always use a dedicated front matter field called `subtitle` or `description`.)

So I propose adding a `summary` variable to resources…the return value of which would be determined by the configured summarization service for the site. We could have it return the first paragraph by default if no service is available, just like the current default excerpt behavior. (Except we need to refactor that because the current code which does that is icky!)

jaredcwhite:
Resolves #306 

This new proposed plugin API allows anyone (a site owner, a plugin author) to augment resources with new methods (both via the Liquid drop as well as the standard Ruby base class). Here's an example from the test suite:

```ruby
module TestResourceExtension
  def self.return_string
    "return value"
  end

  module LiquidResource
    def heres_a_liquid_method
      "Liquid #{TestResourceExtension.return_string}"
    end
  end

  module RubyResource
    def heres_a_method(arg = nil)
      "Ruby #{TestResourceExtension.return_string}! #{arg}"
    end
  end
end

Bridgetown::Resource.register_extension TestResourceExtension
```

Now in any Ruby template or other scenario, you can call `heres_a_method` on a resource:

```ruby
@site.resources.first.heres_a_method
```

Or in Liquid, it'll be available through the drop:

```liquid
{{ site.resources[0].heres_a_liquid_method }}
```

The extension itself can be any module whatsoever, doesn't matter — as long as you include a sub-module of `RubyResource` and optionally `LiquidResource`.

----

This PR also contains a way of providing resource summarization. The `summary` method is now available for resources. By default the first line of content is returned, but any resource extension can provide a new way to summarize resources.

```ruby
module TestSummaryService
  module RubyResource
    def summary_extension_output
      "SUMMARY! #{content.strip[0..10]} DONE"
    end
  end
end

Bridgetown::Resource.register_extension TestSummaryService
```

This could do all sorts of wild analysis, AI stuff, call out to a 3rd-party API (and hopefully cache the results somehow), whatever.

Let me know what you think!

### TODO:

- [x] Tests
- [x] Documentation


ayushn21:
Love it ... looks good to me! ..... If an extension needed to be restricted to a certain resource type, say a `post`, that could easily be done by just inspecting the `collection` on the resource right?

jaredcwhite:
> Love it ... looks good to me! ..... If an extension needed to be restricted to a certain resource type, say a `post`, that could easily be done by just inspecting the `collection` on the resource right?

I think so, something like:

```ruby
return unless collection.label == "posts"
```

ayushn21:
Yeah that's pretty much exactly what I was thinking too 👍

PR #347, opened 2021-07-08
Update macOS documentation to refer to zsh instead of bash

collindonnell:
Apple switched the default shell to zsh from bash in macOS 10.15 (Catalina), and
so instructions to update `~/.bash_profile` are incorrect. This change updates
the documentation to instead use `~/.zprofile`.


ayushn21:
Thanks @collindonnell :)

Issue #356, opened 2021-08-03
Tailwind doesn't work when we use yarn deploy


work-leonid:
This error has already been [discussed](https://github.com/bridgetownrb/bridgetown/issues/333). But the error is still there. 

The problem is caused by the postcss plugin [postcss-focus-within](https://github.com/csstools/postcss-focus-within). Bridgetown uses version 3.0.0 of this plugin. In April there was an update of this package to 4.0.0. With a fix "Fixed: Allow :focus-within to appear escaped in a selector".

I suggest to either update the plugin inside of Bridgetown. Or when installing Tailwind add in `package.json` 
```
"resolutions": {
    "postcss-focus-within": "^4.0.0"
  }
```
That way we'll be using the up-to-date version of the plugin and the error will disappear.

ayushn21:
Honestly, I'm a little reluctant to fix something I consider to be Tailwind's problem within Bridgetown. I totally get your intent with this change and it definitely makes for a smoother DX in the short term.

But in my view, we're adding the Tailwind package, not the `postcss-focus-within` package. Tailwind should be pulling in the correct versions of the dependencies it needs to work. 

I'm not a Tailwind user so I wonder if you have any idea about how widespread this issue is outside of a Bridgetown context and what the "official" workaround or fix for it is? Based on the official Tailwind recommendation we could figure a way to proceed.

work-leonid:
I think this is not a tailwind problem. According to this  [discussion](https://github.com/csstools/postcss-focus-within/issues/3)  this is a postcss problem. Or maybe even `postcss-preset-env`.

ayushn21:
Yeah you're right, it's a `postcss-preset-env` problem. There's been a PR open for over a year to update it <sigh>
https://github.com/csstools/postcss-preset-env/pull/172

Let's sort this out within Bridgetown.

We need to add the the block you posted above to our `package.json.erb` when the `use-postcss` option is set. And we need to run `yarn set resolution postcss-focus-within:^4.0.0` in the `bridgetown webpack enable-postcss` command.

I can put together a PR later this week, or if you fancy giving it a go, you're more than welcome :)

work-leonid:
>I can put together a PR later this week

Please. To be honest, I don't fully understand what needs to be done and how to do it. :)
Thank you!

PR #366, opened 2021-08-16
Resolve postcss-focus-within to v4

ayushn21:
There's a bug in `postcss-focus-within` v3 which has been fixed in v4; but `postcss-present-env` still relies on v3. There's a PR open to update it for over a year: https://github.com/csstools/postcss-preset-env/pull/172

To provide a better developer experience, I reckon it's best to fix this within Bridgetown by explicitly resolving this package to v4. I've set the resolution in our default `package.json`. I don't think there's much point in adding this resolution just when the `--use-postcss` flag is added as it doesn't really do any harm and I don't want to litter the file with multiple `if`s.

Also, this approach means we don't have to set the resolution in our `bridgetown webpack enable-postcss` command which makes life easier as the `yarn set resolution` command only exists in Yarn 2.

Fixes https://github.com/bridgetownrb/bridgetown/issues/356

### DIFF
--- a/bridgetown-core/lib/site_template/package.json.erb
+++ b/bridgetown-core/lib/site_template/package.json.erb
@@ -33,5 +33,8 @@
     "webpack-cli": "^4.7.2",
     "webpack-manifest-plugin": "^3.1.1",
     "webpack-merge": "^5.8.0"
+  },
+  "resolutions": {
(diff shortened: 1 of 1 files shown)


ayushn21:
Hey @jaredcwhite, could you please take a quick peek at this so we can merge it? Thanks! :)

Issue #368, opened 2021-08-19
Empty lines on package.json


eclectic-coding:
When creating a new project: `bridgetown new my_awesome_site`, there are blank lines in the `devDependencies` section of the `package.json`. This does not affect the performance or stop the server, it is just an anomaly. 

**Bridgetown Version**: 0.21.3

**Current behavior**

```json
"devDependencies": {
    "browser-sync": "^2.26.7",
    "concurrently": "^5.2.0",
    "css-loader": "^4.3.0",
    "esbuild": "^0.12.7",
    "esbuild-loader": "^2.13.1",
    "file-loader": "^6.2.0",
    "mini-css-extract-plugin": "^1.3.1",
    
    "sass": "^1.32.8",
    "sass-loader": "^8.0.2",
    
    "webpack": "^5.39.1",
    "webpack-cli": "^4.7.2",
    "webpack-manifest-plugin": "^3.1.1",
    "webpack-merge": "^5.8.0"
  }
```

**Additional context**
The blank lines are probably caused by the conditionals found in [json.erb](https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/site_template/package.json.erb):
```erb
"devDependencies": {
    "browser-sync": "^2.26.7",
    "concurrently": "^5.2.0",
    "css-loader": "^4.3.0",
    "esbuild": "^0.12.7",
    "esbuild-loader": "^2.13.1",
    "file-loader": "^6.2.0",
    "mini-css-extract-plugin": "^1.3.1",
    <% if options["use-postcss"] %>
    "postcss": "^8.3.0",
    "postcss-flexbugs-fixes": "^4.1.0",
    "postcss-loader": "^4.3.0",
    "postcss-preset-env": "^6.7.0",
    <% else %>
    "sass": "^1.32.8",
    "sass-loader": "^8.0.2",
    <% end %>
    "webpack": "^5.39.1",
    "webpack-cli": "^4.7.2",
    "webpack-manifest-plugin": "^3.1.1",
    "webpack-merge": "^5.8.0"
  }
```
**Computing environment (please complete the following information):**
 - OS: Linux/POP_OS/Ubuntu
 - Browser: Firefox
 - Browser Version: 91.0.1
 - Ruby Version: ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]

ayushn21:
[...]

PR #369, opened 2021-08-19
Fix strip out newlines in generated package.json

eclectic-coding:
This is a 🐛 bug fix.

## Summary
Removes empty lines in package.json when creating a new site. 

## Context

  The following will close issue #368 when PR is merged.


ayushn21:
While you're here, could you please make the same change to the conditionals in this file?
https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/site_template/frontend/javascript/index.js.erb

Thanks! :)

PR #374, opened 2021-08-31
add styling to navbar

gnpashi:
This is a 🙋 feature 

 The test suite passes locally (I ran `script/cibuild` to verify this), without adding new tests

## Summary

I have added styles in `site_template/index.scss` and  `site_template/index.css `and changed  `site_template/_components/navbar.liquid` in order to style the navbar and the default layout a little


jaredcwhite:
Thanks for working on this @gnpashi. The default site template is essentially throw away…in no respects do we ever encourage anyone to use it verbatim, so I want it to be as barebones as possible.

It's possible down the road we might ship with a more robust set of themes to choose from (optionally), in which case this type of effort would be most appreciated.

Issue #371, opened 2021-08-28
Investigate a pure Ruby/Rack/Roda(?) implementation of live reloading in development


jaredcwhite:
We're currently using BrowserSync which is a JS/Node tool to provide two services:

* live reloading of the page whenever files change in the repo
* exposing the localhost dev server via proxy to the computer's LAN

I'd like to investigate replacing BrowserSync with a pure Ruby solution, ideally integrated into our new Rack (maybe Roda specifically) based server (see #281).

jaredcwhite:
Maybe based on this: https://github.com/guard/guard-livereload

jaredcwhite:
Now working in #383!

PR #383, opened 2021-09-06
WIP: SSR & file-based dynamic routes in `src/_routes`

### LINKED ISSUE #379: feat: File-based SSR routes
In 2021, it's not enough to be a solely a static site generator. Major Jamstack frameworks are evolving (or have already) to support dynamic routes, often implemented as serverless functions. Increasingly the role of a traditional, separate "backend" is being questioned for many types of websites.

I'm not on the serverless train personally, but I've become convinced of the value of having a static site with choice dynamic routes to handle user presence, e-commerce, and live communication concerns.

In #269 back in April I wrote: "Not trying to compete directly with Sinatra/Rails/etc"

That's true insofar as I have no interest in directly building a Rack-based "backend" web framework.

However, I've done enough work behind the scenes at this point to validate the combination of Bridgetown + Roda. [Roda](http://roda.jeremyevans.net) is a revelation—it supports all the use cases of a "Sinatra" while being able to scale to compete with a "Rails". What Roda is missing however (IMO) is a compelling, out-of-the-box file-based routing and view layer configuration.

That's where we come in. 😎

In the prototype I've been calling "Bridgetown Arch", you can create single files right in `src/_routes` that combine Roda code at the top in a "front matter"-like manner with a resource-based view template below. Something like this:

```md
~~~ruby
# src/_routes/example.md

r.is do
  # Index action
  r.get do
    render_with data: {
      layout: :page,
      title: I18n.t("titles.example"),
      current_time: Time.now,
    }
  end

  # Create action
  r.post do
    puts r.params[:something] # incoming form field!
    flash.info = "Created Sucessfully"
    r.redirect
  end
end
~~~

### Flash: {{ flash.info }}

Hello **folks**! The time is: {{ resource.data.current_time }}

<form action="{{ request.path }}" method="POST">
  {%= csrf_tag(request.path) %}
  <input name="something" />
  <input type="submit" value="Create" />
</form>
```

(Yes, this is a Markdown file + Serbea templating!)

I also have file-based route placeholders working so you could save a file as: `src/_routes/subfolder/[id]/[whatever].erb` and it handles those parameters just as you would expect ,similar to other frameworks like Next or SvelteKit.

I had originally intended to wrap this functionality into a standalone gem literally called `bridgetown-arch`, but upon further reflection I consider this an optional yet entirely valid aspect of core Bridgetown functionality going forward.

So just like the builder and pagination gems within core, I plan to create a new gem (perhaps called `bridgetown-routes`) which sites can opt into using which then adds this file-based dynamic routing—deployable on Render, Fly, or other forward-thinking hosting platforms.

Work done for this can also in some ways benefit backend Rails or other frameworks should people prefer those options.

This directly builds upon the work in #281 as well as addresses #123.

jaredcwhite:
Builds upon PR #281 and resolves #379.

Lots to go on this but early results are *very* promising…


jaredcwhite:
I've added a basic "live reload" in this PR and removed BrowserSync to address #371. It uses polling which initially seems bad but I've actually had real issues with websockets and livereload connections not working over some remote dev environments, so an integration via normal http isn't all that problematic IMHO.

jaredcwhite:
Note to self: fix failing tests before final merge (duh)

bridgetown-core/lib/site_template/Gemfile.erb
```diff
@@ -17,3 +17,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 # Happy Bridgetowning!
 
 gem "bridgetown", "~> <%= Bridgetown::VERSION %>"
+
+# Puma is a Rack-compatible server
+# (you can optionally limit this to the "development" group)
+gem "puma", "~> 5.2"
```
ayushn21, line 23: Maybe it's worth putting this in `development` by default? I'd imagine the vast majority of people would use it only use it locally.

ayushn21 (approved):
Awesome stuff. That was a mammoth of a PR :D. 

I can't say I understood everything but there was nothing bad that jumped out at me. Looking forward to having this in `main`!

Do you think it's worth duplicating `bin/bridgetown` to `bin/bt`? It's a tiny amount of code that's unlikely to change so duplicating the file isn't the big deal in my opinion. It'll avoid putting the onus on the developer to set up an alias and provide a concise way to invoke commands out of the box.

ayushn21:
Just thinking out loud, I have no idea if this is even feasible. 

For live reloading, I wonder if it'd be possible to use `EventSource` to write a message to the client to refresh itself? We'd need to create the connection to a `live_reload` endpoint and listen for messages; that bit is fairly trivial. We'd also need to have some kind of file watcher and write a message to the client when a file changes. Not sure how feasible that is.

Seems like it'd be an easier way to go about it than using websockets; and also a bit more technically "correct" I suppose since the client never needs to write anything in a live reloading scenario.

jaredcwhite:
> For live reloading, I wonder if it'd be possible to use `EventSource` to write a message to the client to refresh itself? We'd need to create the connection to a `live_reload` endpoint and listen for messages; that bit is fairly trivial. We'd also need to have some kind of file watcher and write a message to the client when a file changes. Not sure how feasible that is.

TIL about EventSource!

I'm totally open to any push-based mechanism in addition to polling, or instead of (if it's reliable enough over non-localhost connections).

bridgetown-core/lib/site_template/Gemfile.erb
jaredcwhite: I guess the question is which DX is worse: installing the puma gem in production even if it's not required for a static site build, or _not_ installing it even when someone might need to utilize SSR? I tend to think the latter would be annoying and we'd have to document it: "hey, wanna do SSR? you need to move puma out of development in the Gemfile…" hmm

jaredcwhite:
> Do you think it's worth duplicating `bin/bridgetown` to `bin/bt`? It's a tiny amount of code that's unlikely to change so duplicating the file isn't the big deal in my opinion. It'll avoid putting the onus on the developer to set up an alias and provide a concise way to invoke commands out of the box.

I like that idea!

bridgetown-core/lib/site_template/Gemfile.erb
ayushn21: Yeah fair point. It's not doing any harm outside the dev group and saves the need to go and look up some docs 👍

ricsdeol:
Hi, this is amazing project,  but I didn't find how I render this `_route/item/[slug].erb` ? 

I tried to use this example:

server/routes/count.rb
```ruby
# Rename this file to hello.rb to try it out

class Routes::Count < Bridgetown::Rack::Routes
  route do |r|
    # route: GET /count
    r.get "hello"  do
      { oi: "friend from server" }
    end
  end
end
``` 

src/_routes/count/hello.liquid
```liquid
<p><strong>Oi:</strong> {{ resource.data.oi }}</p>
```


src/_layouts/default.liquid
```liquid
 {% render "count/hello" %}
``` 
```liquid
 {% render "routes/count/hello" %}
``` 

And I get this error: 

```
[Bridgetown]   Liquid Exception: Liquid error (line 10): No such template 'count/hello' in /home/ricsdeol/projects/leaning/count_teste/src/_layouts/default.liquid
```

Could you please give some examples using data from server ? 
Is the server intended to be accessed only through JavaScript request?

jaredcwhite:
@ricsdeol In case you didn't see my reply on Discord…

I think you don't need any of the examples except just the single `src/_routes/count/hello.liquid` file, and it should look something like this: https://gist.github.com/jaredcwhite/b14285934746202236db0b9bbdf2071a

Issue #365, opened 2021-08-16
feat: Some additional front matter DSL niceties


jaredcwhite:
This is sort of a placeholder issue to investigate some possible DX improvement for the DSL we provide for Ruby front matter. Thoughts:

* It'd be nice to have support for nested data. Like:

```ruby
front_matter do
  title "Hello"
  nest_this do
    title "Hi again!"
  end
end
```

Then you'd have values in `resource.data.title` plus `resource.data.nest_this.title`, etc.

* Convert symbol values to strings automatically (might dovetail into #357). This might prove handy: https://apidock.com/rails/v6.1.3.1/Hash/deep_transform_values

jaredcwhite:
Done!

PR #398, opened 2021-09-24
Enhance front matter DSL with nesting and lambda value eval

jaredcwhite:
Partially addresses #365 

Also adds a neat feature where you can pass an evaluation scope and then use that within a lambda value. For example:

```ruby
special_config = Struct.new(:should_output).new
special_config.should_output = false

front_matter eval_with: special_config do
  title "The Weeknd"
  config do
    output eval: -> { should_output }
  end
end
```

This will resolve to:

```ruby
{
  title: "The Weeknd",
  config: {
    output: false
  }
}
```


jaredcwhite:
Renamed `eval_with:` and `eval: ->` to `scope:` and `from: ->`

PR #405, opened 2021-10-01
Add missing regex

### LINKED ISSUE #399: Missing `$` for Webpack's js rule in default config
The default Webpack config file currently has:

```js
const jsRule = {
  test: /\.(js|jsx)/,
  use: {
    loader: "esbuild-loader",
    options: {
      target: 'es2016'
    },
  },
}
```

This will actually match any file with `.js` in it including `foo.js.txt.bat.png`. We need to ensure it's the file extension only by adding a dollar sign:

```js
const jsRule = {
  test: /\.(js|jsx)$/,
  use: {
    loader: "esbuild-loader",
    options: {
      target: 'es2016'
    },
  },
}
```

nachoal:
This is a 🐛 bug fix. 

## Summary

This adds the missing regex `$` as per this [issue](https://github.com/bridgetownrb/bridgetown/issues/399 ) to avoid matching the wrong strings, e.g. `.js.gif` 

## Context

Fixes #399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-.
+### Fixed
+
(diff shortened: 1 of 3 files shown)


jaredcwhite:
Thanks @nachoal!

PR #441, opened 2021-11-08
Make --version works.

Eric-Guo:
Currently `bridgetown --version` is break.


Eric-Guo:
seems it's harder problem than I thought, have to leave it along here.😂

jaredcwhite:
@Eric-Guo maybe it's just a simple tweak to this line?

https://github.com/bridgetownrb/bridgetown/blob/17304510219717524795794fd8d9bc6e5df387b7/bridgetown-core/bin/bridgetown#L22

PR #442, opened 2021-11-08
No need pin ruby2js here.

Eric-Guo:
it's introduced at #243


jaredcwhite:
Not sure what the purpose of this is.

Eric-Guo:
I check again, found `ruby2js` is required by npm package `webpacker-loader`, anyway no need to pin `regexp_parser` version.

jaredcwhite:
Thanks, looks good to me. :+1:

PR #443, opened 2021-11-08
Show archive state for plugins.

Eric-Guo:
At least can save one click for user


jaredcwhite:
If the repo has been archived, we should probably not show it in the list at all, don't you think @Eric-Guo?

jaredcwhite:
Awesome!

PR #458, opened 2021-11-21
Switch to streaming/event source mechanism for live reload

### LINKED ISSUE #416: Investigate improving live reloading in development
Currently, live reloading works by polling an endpoint that checks for changes and refreshes the page:

https://github.com/bridgetownrb/bridgetown/blob/90a73bb5c74f005dddd90e263424d9872fde450d/bridgetown-core/lib/bridgetown-core/utils.rb#L402
https://github.com/bridgetownrb/bridgetown/blob/90a73bb5c74f005dddd90e263424d9872fde450d/bridgetown-core/lib/bridgetown-core/rack/routes.rb#L56

This works well but I think there's scope for improvement here. 

The first approach worth investigating is [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events). We should be able to set up the `live_reload` endpoint to stream responses using the [Roda Streaming Plugin](http://roda.jeremyevans.net/rdoc/classes/Roda/RodaPlugins/Streaming.html). On the client side we can use [`EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) to create the connection, listen for the events and trigger a refresh. There's still a lot of unknowns like how to watch the files and send responses; or whether this approach will even work.

If the above doesn't work, it's also worth investigating a WebSockets based approach I reckon.

jaredcwhite:
Resolves #416 (maybe?)

Well this certainly ended up being an inordinate amount of work! Getting the streaming mechanism enabled wasn't too bad, but then I ran into a bad situation where Ctrl-C would hang until Puma determined the stream had completed (and the longer it would run, the longer it would hang!). It took quite a while to workaround that, and then I discovered long streams would also gunk up Puma's threads if enough pages were opened at once and you were clicking around, because of a delay between a page change and the socket (browser? Puma?) cleaning up and stopping the stream.

I think it's pretty balanced now, between the length of the stream before a browser-initiated reconnection and the interval checking the modification, as well as further modification checks in-browser between a previous stream and a reconnected stream. Good grief.

I tested in Safari and Firefox…haven't tried any other browsers yet.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/rack/routes.rb
+++ b/bridgetown-core/lib/bridgetown-core/rack/routes.rb
@@ -2,6 +2,12 @@
 
 module Bridgetown
   module Rack
+    @interrupted = false
+
+    class << self
+      attr_accessor :interrupted
+    end
+
     class Routes
       class << self
         attr_accessor :tracked_subclasses, :router_block
@@ -37,7 +43,7 @@ module Bridgetown
         def start!(roda_app)
           if Bridgetown.env.development? &&
               !Bridgetown::Current.preloaded_configuration.skip_live_reload
-            setup_live_reload roda_app.request
+            setup_live_reload roda_app
           end
 
           Bridgetown::Rack::Routes.tracked_subclasses&.each_value do |klass|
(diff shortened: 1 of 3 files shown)


ayushn21:
Damn, sounds like you went down quite the rabbit hole with this one 😬. I'm not sure there's a tangible benefit to this approach for the added complexity though 🤷. I know you mentioned that on Twitter as well, but honestly just for the effort put in, if it works, we should probably go for it and see what happens!!

bridgetown-core/lib/bridgetown-core/rack/routes.rb
```diff
@@ -2,6 +2,12 @@
 
 module Bridgetown
   module Rack
+    @interrupted = false
+
+    class << self
+      attr_accessor :interrupted
```
ayushn21, line 8: Can we use ActiveSupport's `mattr_accessor` here by any chance?

jaredcwhite: Good question…I think `mattr_accessor` is particularly useful when object instances which have included a module want to access that module's accessor easily. No need for that here, so probably wouldn't need the overhead

jaredcwhite:
Having done some work recently on a project without this PR and seeing `live_reload` requests cluttering up the Network tab, I think this PR will be a good step in the right direction.

PR #483, opened 2022-01-15
feat: add bundled configuration for Vercel

andrewmcodes:

## Summary

Adds a bundled configuration for Vercel with correct security headers and content type for feeds.xml (if using the RSS plugin).

It also correctly assigns the Bridgetown URL to the VERCEL_URL in non-production environments. This means that if a VERCEL_URL is in the environment, whether production or running via Vercel's CLI locally, the absolute url's will adjust accordingly.

See example of it in use in the [deploys on this PR](https://github.com/andrewmcodes/bridgetown-vercel-demo/pull/1)

## Context

I spoke about this with Jared. I think pre-1.0 we should make the deployment path to whatever service of your choosing as simple as possible without any need to dig into why something is behaving odd.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/vercel.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/vercel.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+copy_file in_templates_dir("vercel.json"), "vercel.json"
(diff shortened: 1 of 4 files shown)


jaredcwhite (approved):

jaredcwhite:
Thanks @andrewmcodes!

andrewmcodes:
🎉

PR #516, opened 2022-03-18
Apply default theme for syntax highlighting

### LINKED ISSUE #512: feat: Default theme for syntax highlighting
## Summary

Apply a theme for syntax highlighting out-of-the-box, so code snippets are styled by default.

## Motivation

I installed Bridgetown 1.0.0 following the [guides](https://www.bridgetownrb.com/docs/installation). Fired up the server with `bin/bridgetown start` and was impressed to see the site, homepage and posts. I viewed the post and saw the mention of `powerful support for code snippets` but was disappointed to see the code snippet was not styled. 

bridgetown-v1-code-snippets

At first I thought I'd done something wrong. The implication of `powerful support` implies that it handles the colour styling for you. I tried to look through the docs and found a reference to the [liquid syntax](https://www.bridgetownrb.com/docs/liquid/tags#code-snippet-highlighting) eg. `{% highlight ruby %}`. Tried that but it made no difference. I wanted to live in markdown syntax anyway and prefer eg. ```ruby

At this point I felt a bit lost and started randomly googling. Eventually I found the information I needed hidden in the Liquid Tags docs page [Stylesheets for syntax highlighting](https://www.bridgetownrb.com/docs/liquid/tags#stylesheets-for-syntax-highlighting). My mental model didn't associate this with Liquid which is why I didn't really explore this section of the documentation before. I do wonder if the syntax highlighting section could be lifted up out of the Liquid docs and placed elsewhere for better visibility?

Once I found these docs it was easy. I understood that there are many styles and I can easily copy one of the linked themes and use it.

bridgetown-syntax-highlighting-styled

But I still think having a default theme for code snippets would better meet new users' expectations. The default application already has some default styles. As long as we make it easy to change I think having some default is a better 'new app' experience than no styles at all. I believe this matches one of Bridgetown's core principles of [Convention over configuration](https://www.bridgetownrb.com/docs/philosophy/#core-principles).

## Syntax highlighting

Code snippets are styled using a `syntax-highlighting.css` stylesheet in your `styles` directory (these styles are imported via your `index.js` file). By Default we use Github's theme, but you can use any CSS that's compatible with Pygments—example gallery [here](https://jwarby.github.io/jekyll-pygments-themes/languages/ruby.html) or from its [repository](https://github.com/jwarby/jekyll-pygments-themes).

````
```ruby
def print_hi(name)
  puts "Hi, #{name}"
end
print_hi('Tom')
#=> prints 'Hi, Tom' to STDOUT.
```
````

Becomes

syntax-highlighting-styled

This section probably needs fleshing out but it's a start.

As an implementation suggestion I think a separate css file for syntax highlighting makes sense to encapsulate this in one place. Here's an example setup app (what I ended up with):

bridgetown-syntax-files-suggestion

We want to make it clear how they can can change the styles and I had a go at adding some code comments (seen in the screenshot on the right).

## Drawbacks

There may be some users that don't want syntax highlighting, or they want a different theme. I still feel _most_ users would expect some theme by default. To cater to all users we need to make it easy to change / remove.

## Unresolved Questions

Which theme to use? For lack of a better option I might vote for Github's theme as it's well known. But I don't have strong opinions here.

### PR DESCRIPTION
**Pull request by jaredlt**

## Summary

This change adds a default theme for syntax highlighting so users
can see the full power of syntax highlighting when they create a
new bridgetown app.

The theme is stored in a new `syntax-higlighting.css` files.

## Context

- Closes #512
- This is a draft PR to promote discussion on this potential enhancement. I know there are further discussions happening in the issue, but perhaps this could be a good first step, or perhaps it will just help aid the ongoing conversation.
- This PR is unfinished, but I wanted to get something up and get feedback first. The things I know are missing are:
  - Upgrade path
  - Changelog
  - Documentation eg. perhaps extract Syntax Highlighting out of [liquid tags docs](https://www.bridgetownrb.com/docs/liquid/tags#stylesheets-for-syntax-highlighting) to its own page and use some of the content from the Guide-level explanation of the issue

### Before

no-theme-code-snippet

### After

default-theme-code-snippet

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/new.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/new.rb
@@ -166,11 +166,14 @@ module Bridgetown
 
       def configure_sass
         copy_file("frontend/styles/index.css", "frontend/styles/index.scss")
+        copy_file("frontend/styles/syntax-highlighting.css",
+                  "frontend/styles/syntax-highlighting.scss")
       end
 
       def configure_postcss
         template("postcss.config.js.erb", "postcss.config.js")
         copy_file("frontend/styles/index.css")
+        copy_file("frontend/styles/syntax-highlighting.css")
       end
 
       # After a new site has been created, print a success notification and
--- a/bridgetown-core/lib/bridgetown-core/commands/new.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/new.rb
@@ -116,6 +116,7 @@ module Bridgetown
         template("frontend/javascript/index.js.erb", "frontend/javascript/index.js")
         template("src/index.md.erb", "src/index.md")
         template("src/posts.md.erb", "src/posts.md")
+        copy_file("frontend/styles/syntax-highlighting.css")
 
         case options["templates"]
         when "erb"
@@ -166,14 +167,11 @@ module Bridgetown
 
       def configure_sass
         copy_file("frontend/styles/index.css", "frontend/styles/index.scss")
-        copy_file("frontend/styles/syntax-highlighting.css",
-                  "frontend/styles/syntax-highlighting.scss")
       end
 
       def configure_postcss
         template("postcss.config.js.erb", "postcss.config.js")
         copy_file("frontend/styles/index.css")
-        copy_file("frontend/styles/syntax-highlighting.css")
       end
 
       # After a new site has been created, print a success notification and
--- a/bridgetown-core/lib/site_template/frontend/javascript/index.js.erb
+++ b/bridgetown-core/lib/site_template/frontend/javascript/index.js.erb
@@ -1,7 +1,9 @@
 <%- if postcss_option -%>
 import "index.css"
+import "syntax-highlighting.css"
 <%- else -%>
 import "index.scss"
+import "syntax-highlighting.scss"
 <%- end -%>
 
 // Import all JavaScript & CSS files from src/_components
--- a/bridgetown-core/lib/site_template/frontend/javascript/index.js.erb
+++ b/bridgetown-core/lib/site_template/frontend/javascript/index.js.erb
@@ -1,10 +1,9 @@
 <%- if postcss_option -%>
 import "index.css"
-import "syntax-highlighting.css"
 <%- else -%>
 import "index.scss"
-import "syntax-highlighting.scss"
 <%- end -%>
+import "syntax-highlighting.css"
 
 // Import all JavaScript & CSS files from src/_components
 <%- if frontend_bundling_option == "esbuild" -%>
--- a/bridgetown-core/lib/site_template/frontend/styles/syntax-highlighting.css
+++ b/bridgetown-core/lib/site_template/frontend/styles/syntax-highlighting.css
@@ -1,0 +1,77 @@
+/* 
+  Syntax Highlighting for Code Snippets
+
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by ayushn21** [comment]
This is a great first step @jaredlt, thanks for the PR :)

To solve the upgrade path issue and perhaps improve the developer experience, I wonder if we could write a simple [Bundled Configuration](https://www.bridgetownrb.com/docs/bundled-configurations) to install themes? Rouge has a number of themes included that can be rendered to CSS using Ruby as such:

```ruby
Rouge::Theme.find('base16.light').render(scope: '.highlight')
```

I'm thinking we could add a configuration so we can run the following command:

```shell
$ bin/bridgetown configure syntax-theme github
```

That would render the GitHub theme from Rouge to CSS as shown above and write it to a `syntax-higlighting.css` file as you've done in this PR.

The in the `new` command all we need to do is call this configuration and it'll setup the GitHub theme by default. It also allows the user to change themes to any other Rouge supported theme by running a single command. And if they want another theme, say from Pygments, they can manually change the CSS file. (We could signpost this in the docs).

You can see examples of bundled configurations we already have here: https://github.com/bridgetownrb/bridgetown/tree/main/bridgetown-core/lib/bridgetown-core/configurations. If you create a new Ruby file there, Bridgetown will automatically pick it up, no extra config necessary.

Here's an example of the `configure` command being invoked from the `new` command: https://github.com/bridgetownrb/bridgetown/blob/8e92eaedddd8ace49d3e1ff96e21bf2182e92aec/bridgetown-core/lib/bridgetown-core/commands/new.rb#L191

I think this achieves a good balance between simplicity and flexibility. What do you think?

jaredlt:
@ayushn21 I love it!

If you're happy for me to carry on I'll have another dig into this and see what I can do. I should be able find some more time next Friday.

ayushn21:
Absolutely, go for it 👍. Drop me a line if you have any questions or you get stuck or something :)

jaredlt:
Time totally got away from me today but I do plan on tacking another look at this next week.

Random thought: I can't decide if having a destructive configuration command is a bad thing or not. If users are just using the command it's fine, but if users customize and then use the command it will overwrite their styles. Assuming everyone has their app under git it should be fine (and if someone has customized the styles and then used the command we can probably assume they know what they're doing), but still, I feel a little nervous about it. Am I over thinking it? What do you think?

jaredcwhite:
@jaredlt If we use Thor actions correctly, there shouldn't be a risk of destruction. If they'd previously created/edited a `syntax-highlighting.css` file, it would prompt the user before overwriting it.

ayushn21:
What Jared said ... also I think it's a safe assumption that everyone would have their site under git ... our `webpack` command overwrites the webpack config file because I assumed everyone would be using git and could just discard the changes if they wanted ...

jaredlt:
@jaredcwhite sounds good! Apologies for my lack of involvement with moving this forward recently. Life and work have been pretty busy :/ I still plan to work through https://github.com/bridgetownrb/bridgetown/pull/534 and get it in a good state, but not sure how much time I'll have the next few weeks. If someone else wants to pick it up I'll hold no grudge :) but equally I do intend to get back to it.

jaredcwhite:
No worries. It'll happen when it happens. ☺️

jaredcwhite:
@jaredlt I simplified things just a tad as both the Sass and PostCSS variants can import the same file extension. Thanks again for the PR!

PR #534, opened 2022-04-14
Add syntax-theme configuration command

jaredlt:

## Summary

This change introduces a new bundled configuration `syntax-theme`
which sets a syntax highlighting theme for code snippets.

To specify _which_ theme to use, this change introduces the concept
of a configuration option which is set by defining the option after
the configuration, separated by a colon `:`. Eg.

`bridgetown configure syntax-theme:github` (`github` is the option)

We use the themes from the Rouge gem to output the required CSS.

image

## Context

- Closes https://github.com/bridgetownrb/bridgetown/issues/512
- Closes https://github.com/bridgetownrb/bridgetown/pull/516 (I opened a new PR to play with this new approach)

I wanted to push this up to get your early feedback. There are still many things to do:

- `bridgetown/bridgetown-core/test/fixtures/test_configuration_with_option.rb:7: warning: instance variable @configuration_option not initialized` when run tests bridgetown-core/test/test_configure_command.rb
- Detect and handle css vs scss
- Call syntax-theme configuration when running `bridgetown new`
- Add `import "syntax-theme.css"` to index.js as part of configuration (ensure it doesn't re-add it if already exists)
- Documentation
- Update `syntax-theme.css` header comments when documentation is finalised
- Test "perform multiple configurations when option provided" doesn't work
- conflict Overwrite UX confusion? (because user is being asked to overwrite a file which is scary (and they didn't create it in the first place). Should we say something in the output ahead of this? Or cover it in the documentation? Both?) Eg. what happens now:

```bash
$ bin/bridgetown configure syntax-theme:molokai
syntax_theme  Generating syntax theme for molokai
    conflict  frontend/styles/syntax-theme.css
Overwrite /home/jaredt/projects/bridgetest/frontend/styles/syntax-theme.css? (enter "h" for help) [Ynaqdhm]
```

Sorry for my slowness on this. I thought I had time to look at this weeks ago but life got in the way! Would love your feedback and then happy to carry on if you agree with the direction.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/configure.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/configure.rb
@@ -38,10 +38,18 @@ module Bridgetown
       protected
 
       def configure(configuration)
-        configuration_file = find_in_source_paths("#{configuration}.rb")
+        if configuration.include?(":")
+          configuration_filename = configuration.split(":").first
+          configuration_option = configuration.split(":").last.split.first
+        else
+          configuration_filename = configuration
+        end
+
+        configuration_file = find_in_source_paths("#{configuration_filename}.rb")
 
         inside(New.created_site_dir || Dir.pwd) do
-          @templates_dir = File.expand_path("../configurations/#{configuration}", __dir__)
+          @templates_dir = File.expand_path("../configurations/#{configuration_filename}", __dir__)
+          @configuration_option = configuration_option
           apply configuration_file, verbose: false
         end
       end
--- a/bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# TODO: handle if css vs scss
+
+theme = @configuration_option
+valid_themes = Rouge::Theme.registry.map { |k, _v| k }.sort
+
+unless valid_themes.include?(theme)
+  @logger.error "\nTheme not valid. Please provide a valid theme
+                   eg. #{"syntax-theme:github".bold.white}"
+  @logger.info "\nValid themes are: \n\n#{valid_themes.join("\n")}\n"
+
+  return
+end
+
+say_status :syntax_theme, "Generating syntax theme for #{theme.bold.white}"
+
+create_file "frontend/styles/syntax-theme.css" do
+  header_comments = <<~CSS
+    /*#{" "}
+      Syntax Highlighting Theme for Code Snippets
+
+      https://www.bridgetownrb.com/docs/liquid/tags#stylesheets-for-syntax-highlighting
+
+      Other styles available eg. https://github.com/jwarby/jekyll-pygments-themes
+
+      To use another style, run `bin/bridgetown configure syntax-theme:{theme}` from
+      the command line. Eg. `bin/bridgetown configure syntax-theme:github`. To see a
+      list of all available styles run `bin/bridgetown configure syntax-theme`.
+      Or create your own by editing this file manually!
+    */
+
+  CSS
+
+  header_comments +
+    Rouge::Theme.find(theme).render(scope: ".highlight")
+end
--- a/bridgetown-core/test/test_configure_command.rb
+++ b/bridgetown-core/test/test_configure_command.rb
@@ -10,11 +10,13 @@ class TestConfigureCommand < BridgetownUnitTest
   context "the configure command" do
     setup do
       FileUtils.cp(test_dir("fixtures", "test_automation.rb"), configurations_path)
+      FileUtils.cp(test_dir("fixtures", "test_configuration_with_option.rb"), configurations_path)
       @cmd = Bridgetown::Commands::Configure.new
     end
 
     teardown do
       File.delete("#{configurations_path}/test_automation.rb")
+      File.delete("#{configurations_path}/test_configuration_with_option.rb")
     end
 
     should "list all available configurations when invoked without args" do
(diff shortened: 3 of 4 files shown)


bridgetown-core/lib/bridgetown-core/commands/configure.rb
```diff
@@ -38,10 +38,18 @@ def self.source_root
       protected
 
       def configure(configuration)
-        configuration_file = find_in_source_paths("#{configuration}.rb")
+        if configuration.include?(":")
+          configuration_filename = configuration.split(":").first
+          configuration_option = configuration.split(":").last.split.first
```
jaredlt, line 43: The final `split.first` is to ignore any text _after_ the option. It seems if you run multiple configurations in a single command the second command shows up?

Eg. `bridgetown configure syntax-theme:github vercel`

`configuration_option` will be `github vercel` without the final `split.first`

It does seem odd to me that this occurs as I thought the `args.each do |configuration|` line above would already sort this?

bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# TODO: handle if css vs scss
```
jaredlt, line 3: I could do with some guidance on how to detect and handle css vs scss here.

bridgetown-core/test/test_configure_command.rb
```diff
@@ -52,5 +54,49 @@ def configurations_path
...
+      capture_stdout do
+        @cmd.invoke(:perform_configurations, ["test_configuration_with_option:github test_automation"])
+      end
+
+      frontend_styles_path = File.expand_path(root_dir("frontend", "styles"), __dir__)
+      assert File.exist?("#{frontend_styles_path}/syntax-test.css")
+      # TODO: This test doesn't work, but manual testing seems to prove it does work fine
+      # I _think_ it's something to do with capture_stdout not capturing everything... maybe?
+      # assert_match %r!fixture.*?Works\!!, output
+      # assert_match %r!create.*syntax-test.css!, output
```
jaredlt, line 81: Couldn't get this to work?

bridgetown-core/test/test_configure_command.rb
```diff
@@ -52,5 +54,49 @@ def configurations_path
...
+      assert_match %r!Theme not valid!, output
+    end
+
+    should "show error when option is invalid" do
+      output = capture_stdout do
+        @cmd.invoke(:perform_configurations, ["test_configuration_with_option:invalid"])
+      end
+
+      assert_match %r!Theme not valid!, output
+    end
```
jaredlt, line 100: These tests were useful to me to ensure it was working well, but technically it's testing stuff within the test file as opposed to running the command, so not sure how valuable they are?

bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
ayushn21: SCSS is a superset of CSS, right? So could we just use plain old CSS all the time?

bridgetown-core/lib/bridgetown-core/commands/configure.rb
ayushn21: Yeah that is quite weird; although I just tried it and it seems to be working as expected for me 🤷‍♂️. i.e. the second configuration option is not passed through to the `configure` method. Could you try again to confirm this is still happening for you?

ayushn21:
```suggestion
      def configure(configuration)
        filename, *options = configuration.split(":")
        file = find_in_source_paths("#{configuration_filename}.rb")

        inside(New.created_site_dir || Dir.pwd) do
          @templates_dir = File.expand_path("../configurations/#{filename}", __dir__)
          @options = options
          apply file, verbose: false
        end
      end
```

bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
```diff
@@ -0,0 +1,37 @@
...
+
+theme = @configuration_option
```
ayushn21, line 5:
```suggestion
theme = @config_options&.first || "github"
```

bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
```diff
@@ -0,0 +1,37 @@
...
+      the command line. Eg. `bin/bridgetown configure syntax-theme:github`. To see a
+      list of all available styles run `bin/bridgetown configure syntax-theme`.
+      Or create your own by editing this file manually!
+    */
+
+  CSS
+
+  header_comments +
+    Rouge::Theme.find(theme).render(scope: ".highlight")
+end
```
ayushn21, line 37: Could potentially do something like this to import the syntax into the css bundle:

```ruby
css_index = Dir["#{Dir.pwd}/frontend/styles/index.*css"].first
prepend_to_file css_index do
  "@import 'syntax-theme.css';"
end
```

That should work for both CSS and SCSS I reckon

bridgetown-core/test/test_configure_command.rb
ayushn21: Yeah I agree this is testing the fixture rather than the actual command so I don't see any value it in personally.

ayushn21:
Great work on this @jaredlt! Thanks for the effort. I've left some comments and suggestions. Let me know what you think :)

bridgetown-core/test/test_configure_command.rb
ayushn21: For these tests I'd recommend using a very simple test configuration. Similar to `test_automation.rb`. All we want to test is that the options were passed to the configuration file successfully. What the configuration script does with the options is not relevant to this test. Does that make sense?

These tests should be testing the `configure` command in a generic way.

bridgetown-core/test/test_configure_command.rb
ayushn21: When invoking the command in this way, every individual argument should be a separate element in the array. So invoking it as:

@cmd.invoke(:perform_configurations, %w[test_configuration_with_option:github test_automation])

should do the trick!

bridgetown-core/lib/bridgetown-core/configurations/syntax-theme.rb
jaredcwhite: Yeah, should work. Let's use double-quotes rather than single quotes for consistency with other files we generate.

jaredcwhite:
I think at this point we should close this—and maybe choosing between different syntax highlighting options should be in a plugin anyway, rather than in core.

PR #538, opened 2022-04-20
Allow route classes to be prioritized to adjust run order

jaredcwhite:
Previously, classes in `server/routes` would add to Roda's run block in "Zeitwerk" order…roughly alphabetical I'd guess. Now they're added (a) by priority, so you can ensure high priority routes are handled before low priority ones, and (b) alphabetical by class name within a single priority tier.

In addition, we now have Rack/Roda/SSR integration testing in our test suite. 🎉 *Lots* more tests to follow!


jaredcwhite:
Builder priority feature was added as well, and priority logic was extracted to a concern. So now base plugins (converters, generators), builders, and routers all share similar logic. (Hooks are still an odd duck, but maybe there's a way to solve that too down the road.)

PR #542, opened 2022-04-24
Add Sass support for esbuild, update frontend dependency versions

### LINKED ISSUE #526: feat: Sass support with esbuild
Currently Sass is only supported when the Webpack frontend bundler is used. We should provide it as an option for esbuild as well.

jaredcwhite:
Like the title says! Sass files are now supported in esbuild. Unlike Webpack, you can easily use Sass files after the fact, it's not mutually exclusive with PostCSS.

Also the versions of the frontend dependencies in package.json have all been updated.

Closes #526 

- [x] Documentation


bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
```diff
@@ -106,6 +129,65 @@ const postCssPlugin = (options) => ({
...
+        .map((x) => x.pathname)
+
+      let cssOutput = result.css.toString()
+
+      if (result.sourceMap) {
+        const basedir = process.cwd()
+        const sourceMap = result.sourceMap
+
+        const promises = sourceMap.sources.map(async source => {
+          const daFile = await fs.readFile(fileURLToPath(source), "utf8")
```
jaredcwhite, line 170: should rename this variable =)

PR #546, opened 2022-05-03
feat: add HTML  & XML Inspectors API using Nokogiri

### LINKED ISSUE #536: feat: allow resources/generated pages to be manipulated using DOM-style tooling
I'm in a hurry so I'll flesh this out later, but after a chat with @claviska I came up with an idea for providing a way for plugins to hook into a DOM-style representation for HTML content (thinking we'll use Nokogiri for this). Upon post-render (aka after going through template engines, Markdown, etc.), one or more plugins could access the "DOM" and make changes, and after all plugin transformations we'd serialize back to HTML. Some examples:

```ruby
  def build
    inspect_html do |document|
      meta = document.query_selector("head > meta[name='description']")
      meta[:content] = "A whole new description"
    end

    inspect_html do |document|
      footer = document.query_selector(:footer)
      footer[:class] = "cool-footer-class"
    end

    inspect_html do |document|
      document.query_selector_all("upcase-me").each do |el|
        el.content = el.content.upcase
      end
    end
  end
```

jaredcwhite:
Closes #536.

- [x] Tests
- [x] Documentation: https://bridgetown-beta-pr-546.onrender.com/docs/plugins/html-inspectors


bridgetown-builder/lib/bridgetown-builder/dsl/html_inspectors.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          def query_selector_all(selector)
+            css(selector)
+          end
+        end
+
+        module RunInspectors
+          def self.call(resource, inspectors) # rubocop:disable Metrics/CyclomaticComplexity
+            return resource.output if !inspectors ||
+              !resource.destination&.output_ext&.starts_with?(".htm") ||
```
andrewmcodes, line 20: warning: I don't particularly have a concrete example use case for this and am just thinking out loud

How hard would it be to allow this to work for XML too? I am thinking about sitemaps and app manifests. If the answer is one or two lines of code changed, I could see it being useful to add. ¯\_(ツ)_/¯

bridgetown-website/plugins/builders/inspectors.rb
```diff
@@ -0,0 +1,14 @@
...
+    inspect_html do |document|
+      document.query_selector_all("article h2[id], article h3[id]").each do |heading|
+        heading << document.create_text_node(" ")
+        heading << document.create_element(
+          "a", "#",
+          href: "##{heading[:id]}",
+          class: "heading-anchor"
+        )
+      end
+    end
```
andrewmcodes, line 12: 😍 This is FANTASTIC

bridgetown-website/src/_docs/plugins/html-inspectors.md
```diff
@@ -0,0 +1,65 @@
...
+category: plugins
+---
+
+The HTML Inspectors API, added in Bridgetown 1.1, provides a useful and safe way to manipulate the HTML output of your resources. Safe because instead of using string manipulation, regular expressions, and the like—which is prone to error—you'll be working on real node trees. This is thanks to [Nokogiri](https://nokogiri.org), a Ruby gem which lets you work with a DOM-like API directly on HTML documents.
+
+Bridgetown doesn't directly install the Nokogiri gem, so be sure to run `bundle add nokogiri` if it isn't already part of your bundle.
+
+## Your First Inspector
+
+Let's add a oft-requested feature to our site: automatic `target="_blank"` attributes on all outgoing links. It's easy with an HTML inspector.
```
andrewmcodes, line 14:
```suggestion
Let's add an oft-requested feature to our site: automatic `target="_blank"` attributes on all outgoing links. It's easy with an HTML inspector.
```

Side note, should `Inspector` be capitalized? My thinking here is that `HTML inspector` is not something you can google for (idts anyways after a quick search), so there may be value in signifying it is a noun related to a feature in this project and not an HTML spec.

bridgetown-website/src/_docs/plugins/html-inspectors.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+You can see this in action right on this very page!
+
+## Performance Considerations
+
+All resources which result in HTML output (rather than JSON or some other format) will be procssed through any defined inspectors. For greater performance and fidelity, the Nokogiri document for a single resource will be the same across all inspectors (rather than instantiating a new Nokogiri document for each inspector).
+
+{%@ Note type: :warning do %}
+Nokogiri [relies on a C extension](https://nokogiri.org/#guiding-principles_1) which in turn uses `libxml2`, so generally you should see very fast performance unless the number of resources in your project is extremely large.
```
andrewmcodes, line 62:
```suggestion
Nokogiri [relies on a C extension](https://nokogiri.org/#guiding-principles_1) which in turn uses `libxml2`. You should see fast performance unless the number of resources in your project is monstrous.
```

Just a suggestion 🙂

bridgetown-website/src/_docs/plugins/html-inspectors.md
```diff
@@ -0,0 +1,65 @@
...
+{% end %}
+
+If you find yourself needing to bypass inspectors for certain resources and avoid the overhead of using Nokogiri at all for those resources, you can set the [front matter](/docs/front-matter) variable `bypass_html_inspectors: true`. (And if you want, you can use [front matter defaults](/docs/content/front-matter-defaults) to apply this to a whole array of resources.)
```
andrewmcodes, line 65:
```suggestion
If you find yourself needing to bypass inspectors for certain, large resources to avoid the overhead of using Nokogiri, you can set the [front matter](/docs/front-matter) variable `bypass_html_inspectors: true` to instruct Nokogiri not to parse that resource. To apply this to a whole array of resources, make it a default with [front matter defaults.](/docs/content/front-matter-defaults)
```

bridgetown-website/src/_docs/plugins/html-inspectors.md
andrewmcodes: Maybe enormous? Huge?

bridgetown-website/src/_docs/plugins/html-inspectors.md
```diff
@@ -0,0 +1,65 @@
+---
+title: HTML Inspectors
+order: 0
+top_section: Configuration
+category: plugins
+---
+
+The HTML Inspectors API, added in Bridgetown 1.1, provides a useful and safe way to manipulate the HTML output of your resources. Safe because instead of using string manipulation, regular expressions, and the like—which is prone to error—you'll be working on real node trees. This is thanks to [Nokogiri](https://nokogiri.org), a Ruby gem which lets you work with a DOM-like API directly on HTML documents.
```
andrewmcodes, line 8: > Safe because instead of ...

I think the beginning of this sentence needs a little love. What is safe? Why would it not be? etc.

andrewmcodes (commented):
This looks incredible. I have been attempting to do something like this performantly (without JS of course) similar to what 11ty has and this is way better than I had even dreamed. This will open up the possibility to do some _really_ cool transformations now. Definitely an area that plugins can flourish!

I left a few nitpicks in the docs in a hope you may like one or two of the suggestions.

Amazing job on this @jaredcwhite 👏

jaredcwhite:
Thanks @andrewmcodes for taking a look and for your enthusiasm. It been super fun to work on and implement…once again gotta give Mr. Shoelace himself @claviska a big shoutout for coming up with the feature idea as he works on his own docs engine.

I'm about to be OOTO for a few days, but I'll take a close look at your feedback as soon as I can. 🙏

ParamagicDev (approved):
This is sexy and looks great and is a surprisingly small amount of code, great work!

jaredcwhite:
Thanks again @andrewmcodes for the feedback! I'll clean up the first few sentences, and I'll also look into supporting `.xml` extension files as well.

PR #548, opened 2022-05-04
invalid multibyte char (US-ASCII) - Remove emojis?

coorasse:
I stumbled upon this error the first time I tried to run `/bin/bridgetown configure gh-pages`:

```
[...]
in `instance_eval': /Users/coorasse/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/bridgetown-core-1.0.0/lib/bridgetown-core/configurations/gh-pages.rb:7: invalid multibyte char (US-ASCII) (SyntaxError)
[...]
```

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/gh-pages.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/gh-pages.rb
@@ -4,10 +4,10 @@
 copy_file in_templates_dir("gh-pages.yml"), ".github/workflows/gh-pages.yml"
 
 # rubocop:disable Layout/LineLength
-say "🎉 A GitHub action to deploy your site to GitHub pages has been configured!"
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Shouldn't need to remove emojis from the codebase. Any ideas @ayushn21?

andrewmcodes:
@coorasse can you run `echo ${LANG-none}` in your terminal and let us know the output? 

If I recall, macOS does not set `$LANG` be default so I am guessing it's an encoding issue with your terminal and that command above will output `none`. If that is the case, add `export LANG=en_US.UTF-8 ` to the top of your `~/.zshrc`. Quit your terminal completely and then reopen and try running the command again. My bet is it will be fine now.

coorasse:
Yes. I confirm that this fixed the problem. I close this, as it will stay as a reference if someone has the same issue in the future. Thanks!

PR #551, opened 2022-05-11
Add Markdown support to convert `==` / `::` to `<mark>`

jaredcwhite:
Many Markdown editors allow you to highlight text with `::double colons::` or with `==double equals==`. I've grown accustomed to this myself in various writing contexts. Because of that, it's frustrating to have to switch gears in Bridgetown `.md` files and add `<mark>` tags everywhere instead.

This PR patches Kramdown's GFM parser (our default) to add support for both those delimiters. So for example:

```md
This is ==now highlighted==! ::Yay!::
```

becomes

```html
This is <mark>now highlighted</mark>! <mark>Yay!</mark>
```

The `kramdown: mark_highlighting` YAML config option can be set to `false` to disable this new feature if for some reason it's a breaking change for existing websites.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configuration.rb
+++ b/bridgetown-core/lib/bridgetown-core/configuration.rb
@@ -83,6 +83,7 @@ module Bridgetown
         "footnote_nr"             => 1,
         "show_warnings"           => false,
         "include_extraction_tags" => false,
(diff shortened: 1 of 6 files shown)


andrewmcodes:
Was just starting to build a regex to do this so 🎉  Great addition! 🎉

PR #563, opened 2022-05-25
Update site template configuration file with link to config options docs

andrewmcodes:

## Summary

This commit adds a link to the configuration options docs directly into the default `bridgetown.config.yml`. This will make navigating to the docs quicker and will also increase discoverability of those options for folks getting started with Bridgetown.

## Context

I often need to reference the configuration options docs when updating my site config. Typically, I have to navigate to the Bridgetown site, and then search/find the link for the configuration options, which takes several clicks and can break my concentration. This change will also ensure that the options are easily discoverable in case someone has trouble finding them on the website.

Not a necessity, but it would be nice to have and provide a smoother DX after running `bridgetown new`.

### DIFF
--- a/bridgetown-core/lib/site_template/bridgetown.config.yml
+++ b/bridgetown-core/lib/site_template/bridgetown.config.yml
@@ -3,6 +3,9 @@
 # This config file is for settings that affect your whole site, values
 # which you are expected to set up once and rarely edit after that.
 #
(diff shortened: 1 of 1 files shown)


jaredcwhite:
@andrewmcodes Makes total sense, and it'll also add more impetus to keep that docs page updated regularly.

PR #564, opened 2022-05-25
feat: add rb files to the Tailwind content glob

andrewmcodes:

## Summary

This commit updates the Tailwind CSS content glob to also include ruby files when searching for classes in the src folder.

## Context

Components, plugins, helpers, data files, etc. that are written in Ruby may contain Tailwind CSS classes that won't get generated by the JIT because they aren't in the content scope. 

I have been using Tailwind for years but it took me close to 45 minutes before I realized the reason my classes were not generating the right CSS because they were in a Ruby component, not the accompanying erb file. Hopefully this will save people heartache and frustration if they run into the same issue and not make it be something you always need to add when generating a new site with Tailwind.

The one downside to this addition could be a performance hit. The larger the glob, the longer it takes Tailwind to scan for classes. 

Here is why I think that is ok:

- It is very easy to go in and scope down the glob if needed
- You would need to have an enormous amount of files for it to start bringing down perf, plus most of them probably won't be pure Ruby making the hit minimal compared to markdown or ERB
- The places you can write Ruby in your site is increasing, so people are bound to come across this and it can be very frustrating

A small change but hopefully it will at minimum save some frustration/time and prevent the need to Google to figure out what's happening.

Thanks for reading 😄

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/tailwindcss.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/tailwindcss.rb
@@ -21,7 +21,7 @@ run "npx tailwindcss init"
 
 gsub_file "tailwind.config.js", "content: [],", <<~JS.strip
   content: [
-      './src/**/*.{html,md,liquid,erb,serb}',
+      './src/**/*.{html,md,liquid,erb,serb,rb}',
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Quite right! Seems like an oversight. LGTM

PR #573, opened 2022-06-08
Fix migrate-from-webpack process

katafrakt:
This is a 🐛 bug fix.

## Summary

Since SASS is now again supported in 1.1.0 beta2, I wanted to finally update my blog from webpack to esbuild by running `bridgetown esbuild migrate-from-webpack`. However, in the process, I encountered two problems:

* There is a code that check if SASS is used and blocks from migrating if it is - this is now not necessary
* `postcss-load-config` released a new version few days ago with breaking API, which causes generated config to not work. Version pinning to last working versions is needed.


jaredcwhite:
Awesome @katafrakt, thanks for the fix.

PR #577, opened 2022-06-11
Add mention that Postcss is only applied to CSS files

mtomov:

## Summary
Update documentation to mention that Postcss is only applied to CSS files.

## Context

It can be confusing where to add the Tailwind directives when having an `index.scss`, so I thought I'd clarify that PostCSS is not applied to Sass files.

Happy to not merge the PR if this is common knowledge.

Thank you!


jaredcwhite:
Hmm, I don't know if this is actually true — I believe with the esbuild config Sass is run first and then PostCSS gets applied to the compiled output.

mtomov:
Yeah, maybe; For me, placing the tailwind requires in the `sass` files with `esbuild` didn't work. Happy to close.

jaredcwhite:
@mtomov Maybe it's something specific to the Tailwind requires. Feel free to submit a PR that makes a specific note of that in the Tailwind configuration docs. I'll close this more general one.

PR #578, opened 2022-06-11
Clarify how to reference static files on the frontend from Webpack/Esbuild

mtomov:

## Summary

Provide examples for referencing static files with Esbuild and Webpack.

## Context

Esbuild and Webpack have a slightly different strategy for referencing static files that are processed by
them. Provide examples to ease use.

Alternatively, we can include those examples directly in the templates, optionally in a commented-out state.

Feel free to close PR if obvious or incorrect.

Thank you!

### DIFF
--- a/bridgetown-website/src/_docs/frontend-assets.md
+++ b/bridgetown-website/src/_docs/frontend-assets.md
@@ -143,6 +143,29 @@ You can use the `asset_path` Liquid tag/Ruby helper to reference assets within t
 
 will look for `frontend/images/folder/somefile.png`.
 
+### Globbing multiple assets
+
+If you need to import an entire folder through the frontend pipeline for example with images, here's how to do it:
+
+###### Esbuild
+
+```js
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Thanks @mtomov for the submission. Can we move this to a separate section on that page which is related to globbing multiple asset files? I personally don't ever glob-import images through the frontend bundling pipeline, so I'd want to make it clear this is an entirely optimal step. Also, we wouldn't ever want to `console.log` anything that's not truly a debug statement. I think simply referencing the variable in a loop (maybe like `images.forEach(image => image)`) would suffice.

jaredcwhite:
@mtomov Can confirm this does work:

```js
import images from "../images/**/*.{jpg,jpeg,png,svg,gif}"
images.forEach(image => image)
```

mtomov:
Hi @jaredcwhite , indeed `Object.entries(images).forEach(image => image)` works!

PR #586, opened 2022-06-24
Switch from rack-indifferent to Roda's indifferent_params plugin

jaredcwhite:
Ha, I don't get to use the `dependencies` label very often! ;-)


jaredcwhite:
PR resolves this conversation: https://github.com/bridgetownrb/bridgetown/discussions/585

PR #596, opened 2022-07-16
Update `postcss-load-config` to `4.x` and address synchronisation issues (*breaking change*)

### LINKED ISSUE #591: Support `postcss-load-config` >= 4.x
Version 4.x of `postcss-load-config` has a breaking change; it drops the `sync()` method upon which the default ESBuild configuration of Bridgetown relies. 

See https://github.com/postcss/postcss-load-config/pull/234 for more information.

**Bridgetown Version**:

`bridgetown 1.1.0.beta3 "Belmont"`

**To Reproduce**

1. Upgrade `postcss-load-config` to `^4.0.0` in `package.json`. 
2. Run `yarn` in the project.
3. Run `bin/bridgetown start`
4. Compilation fails due to error below

```shell
node esbuild.config.js --watch
[Frontend] /Users/stefan/code/org/project/config/esbuild.defaults.js:180
[Frontend] const postCssConfig = postcssrc.sync()
[Frontend]                                 ^
[Frontend]
[Frontend] TypeError: postcssrc.sync is not a function
```

**Current behavior**

Bridgetown fails to compile the CSS.

**Expected behavior**

Bridgetown compiles the CSS.

**Screenshots**

...

**Computing environment (please complete the following information):**
 - OS: MacOS

**Additional context**

I'd like to submit a PR for it and address it. However, I'm not 100% sure how the ESBuild configuration is set up and works for Bridgetown. Would need to dive a little bit deeper.

stefanvermaas:
This is a 🐛 bug fix (?)

## Summary

As reported in #591, `postcss-load-config` introduced a breaking change in their latest major version update: `4.x`. Instead of loading the configuration in a synchronous way, it's needed to load the configuration in an asynchronous way.

The PR;
- Allows loading the PostCSS configuration under `postcss-load-config > 4.x` again.
- Updates the `postcss-load-config` references from `3.1.4` to `4.0.1`.
- Updates the template for the `lit/esbuild-plugins` to also use the asynchronous API.
- Updates the naming of the "import" function of the PostCSS plugin to align it with the other "import" functions in the `esbuild.defaults.js`.

I'm a bit rusty on my JavaScript, so I'm looking for feedback on the proposed `async/await` changes. I'm also not sure how I could add a test case for this and whether or not I should add a test case for this fix.

I also wonder whether or not you'd want to support both `3.x` and `4.x` versions of the `postcss-load-config` package as I'd need to provide a way to use both versions to get the configuration object.

I tested out things locally with the latest version of `postcss-load-config` and was able to build the Bridgetown site again with the new configuration.

## Context

Fixes #591

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
+++ b/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
@@ -84,7 +84,7 @@ const importGlobPlugin = () => ({
 })
 
 // Plugin for PostCSS
-const postCssPlugin = (options, configuration) => ({
+const importPostCssPlugin = (options, configuration) => ({
   name: "postcss",
   async setup(build) {
     // Process .css files with PostCSS
@@ -98,7 +98,7 @@ const postCssPlugin = (options, configuration) => ({
         load: async filename => {
           let contents = await readCache(filename, "utf-8")
           const filedir = path.dirname(filename)
-          // We'll want to track any imports later when in watch mode:          
+          // We'll want to track any imports later when in watch mode:
           additionalFilePaths.push(filename)
 
           // We need to transform `url(...)` in imported CSS so the filepaths are properly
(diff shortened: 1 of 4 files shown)


jaredcwhite:
Thanks for the PR @stefanvermaas! Apologies for the delay in looking into this. I'll try to review soon.

jaredcwhite:
Thanks for your patience @stefanvermaas. Looking good to me! We'll want to make sure we mark this a breaking change as the `build` function call in `esbuild.config.js` will now return immediately with a promise, rather than run synchronously. If folks have added extra code to run after the build function, they'll have to use `.then`:

```js
build(outputFolder, esbuildOptions).then(() => {
  console.log("build complete")
})
```

PR #610, opened 2022-08-05
Specify production BRIDGETOWN_ENV for Github Pages builds

vvveebs:
This is a 🐛 bug fix.

## Summary

The current GitHub Pages configuration builds the site using the default `development` environment.

This PR updates the relevant configuration to use the `production` environment instead as I think that's always going to be the intention for this workflow, and it matches other configurations such as [Netlify](https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/bridgetown-core/configurations/netlify/netlify.toml#L7).


jaredcwhite:
Great catch!

PR #589, opened 2022-07-01
feat: Add hash attributes functionality to view tag helpers

DRBragg:
This is a 🙋 feature or enhancement.

## Summary

Adds the ability to pass a hash to `link_to` (or call `attributes_from_options` to use on any HTML tag) and have the key-value pairs be listed as HTML attributes.

Example:

```ruby
<%= link_to "Label", "/foo/bar", class: "classes", data: { controller: "testable", action: "testable#test" } %>
```
would return:
```html
<a href="/foo/bar" class="classes" data-controller="testable" data-action="testable#test">Label</a>
```
You also have the ability to pass a hash as a keys value and have the values keys merged with the parent key. So:
```ruby
<%= link_to "Label", "/foo/bar", class: "classes", data: { controller: "testable", action: "testable#test", testable: { target: "test_value", index_value: "1" } } %>
```
would return:
```html
<a href="/foo/bar" class="classes" data-controller="testable" data-action="testable#test" data-testable-target="test_value" data-testable-index-value="1">Label</a>
```
This is very helpful when using stimulus controllers.

If you wanted to use this on a different HTML tag you could do:
```ruby
<div <%= attributes_from_options(class: "classes", data: { controller: "testable", action: "testable#test" }) %>>Content</div>
```
and you'd be left with:
```html
<div class="classes" data-controller="testable" data-action="testable#test">Content</div>
```

Naming is hard so I'm definitely open to suggests (I sorta hate `attributes_from_options` TBH).

Let me know if anything isn't to your liking and I'm happy to adjust!

## Context

Close #537

PR #613, opened 2022-08-16
Use correct ERB syntax in docs

DRBragg:

## Summary

I made a goof in #589. A few of the ERB expressions in the docs for `attributes_from_options` were being evaluated within the template. Adjusting the syntax from  `<%=`  to `<%%=`, so it’s escaped, should fix the issue.

## Context

re: Twitter convo w/ @jaredcwhite


jaredcwhite:
Thanks @DRBragg, I also noticed one other tidbit (need to mark the attributes string as safe or they might get escaped twice).

PR #622, opened 2022-09-11
feat: allow template to use `data` directly instead of `resource.data`

### LINKED ISSUE #527: feat: expose `data` to template directly
Something I've thought about for some time is the verbosity of writing `resource.data.X` all the time. If you know you need to call `resource.data` in numerous places, why not just expose `data` directly? So instead of `resource.data.title`, just write `data.title`. We'd still provide `resource` of course because some methods are on the resource object directly like `relative_url`, `relations`, etc., plus 100% backwards compatibility.

(This would also make the following dynamic route example more concise and logical:)

```erb
---<%
# route: /items/:slug
r.get do
  item_id, *item_sku = r.params[:slug].split("-")
  item_sku = item_sku.join("-")

  render_with data: {
    layout: :page,
    title: "Item Page",
    item_id: item_id,
    item_sku: item_sku
  }
end
%>---

<p><strong>Item ID:</strong> <%= data.item_id %></p>

<p><strong>Item SKU:</strong> <%= data.item_sku %></p>
```

Thoughts?

jaredcwhite:
Resolves #527 (including @johlym's feature request!)

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/converters/liquid_templates.rb
+++ b/bridgetown-core/lib/bridgetown-core/converters/liquid_templates.rb
@@ -89,6 +89,7 @@ module Bridgetown
         payload["paginator"] = document.respond_to?(:paginator) ? document.paginator.to_liquid : nil
         payload["layout"] = @layout ? @layout.to_liquid.merge({ data: @layout.data }) : {}
         payload["content"] = content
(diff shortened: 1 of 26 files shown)


johlym:
This looks fantastic 🌮

PR #627, opened 2022-09-18
Update SSR tests to work with Rack 3

vvveebs:

## Summary

- Adding `require "rack"` because Rack no longer seems to be loaded. (Wasn't sure why this was now needed though 🤔.)
- Remove call to `.first` on the return value from `Rack::Builder.parse_file` - in Rack's latest version, [just the app is returned](https://github.com/rack/rack/pull/1663) rather than an array.


jaredcwhite:
@vvveebs Thanks for tracking that down!

Issue #581, opened 2022-06-15
Replace old-school `options = {}` arguments with `**options` where appropriate


jaredcwhite:
Except for cases where we explicitly expect a standard hash to be supplied as an argument rather than kwarg-style calls, we should standardize around the more modern destructuring syntax.

jaredcwhite:
This will be closed by a PR coming shortly.

PR #628, opened 2022-09-19
Kwargs conversion of many methods

neilvanbeinum:
## Summary

Convert many methods and calls to use keyword arguments instead of hashes.

Methods which take a `Configuration` haven't been converted, such as `Bridgetown::Site.new`, or initializers of the `Converter` classes. I started to attempt this but then ran into difficulty, particularly with the converters. I think it's related to `Configuration` being a `HashWithDotAccess::Hash`. I didn't want to go too deep down the rabbit hole, especially as there's a comment suggesting the [config may be refactored](https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/bridgetown-core/configuration.rb#L6) at some point.

## Context

Supports resolving of #581

### DIFF
--- a/bridgetown-builder/test/test_helpers_dsl.rb
+++ b/bridgetown-builder/test/test_helpers_dsl.rb
@@ -23,10 +23,13 @@ class TestHelpers < BridgetownUnitTest
     setup do
       @site = Site.new(site_configuration)
       @builder = HelpersBuilder.new("HelpersBuilder", @site).build_with_callbacks
-      @resource = Bridgetown::Model::Base.build(self, :posts, "im-a-post.md", {
+      @resource = Bridgetown::Model::Base.build(
+        self,
+        :posts,
+        "im-a-post.md",
         title: "I'm a post!",
         date: "2019-05-01",
-      }).as_resource_in_collection
+      ).as_resource_in_collection
       @erb_view = Bridgetown::ERBView.new(@resource)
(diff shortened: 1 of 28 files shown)


jaredcwhite:
Very cool @vvveebs, I appreciate your attention to detail! I should have time to review this after getting the first beta of v1.2 out the door.

vvveebs:
I've got into a tangle with failing ruby 3 tests. I'll promote this draft PR to ready to review if I can resolve these.

jaredcwhite:
I think we can punt on this and reevaluate for v1.3 or whatever…at this point I don't want to introduce any more major changes before the 1.2 release.

jaredcwhite:
Thanks @neilvanbeinum for working on this. I believe there are a number of places where we wouldn't want to make this change, particularly models and anything interfacing with Roda. I may want to back out of this for the time being and review such changes on a case-by-case basis.

sandstrom:
I agree with @jaredcwhite here.

I think keyword arguments is the way to go for new methods, but I wouldn't introduce them like this.

`**options` is mostly an anti-pattern. For internal APIs, one can convert from `options` to real keyword arguments when working on that specific code.

For public APIs, I'd do it full on, i.e. explicitly say which kwargs we are expecting for each method, and then treat it as a breaking change (since it's a public API).

PR #634, opened 2022-10-09
typo: delimeters => delimiters

alexanderadam:

## Summary

I think this is just a tiny typo. It's called [delimiter](https://www.oxfordlearnersdictionaries.com/definition/english/delimit), right?
Because it's the character that limits strings. Or do I misunderstand something here?
I'm not a native speaker so I might be wrong here.


jaredcwhite:
Thanks!

PR #635, opened 2022-10-10
typo: mergable => mergeable

alexanderadam:
This is a 🔦 tiny naming change.

## Summary

I have the feeling that it's called [mergeable](https://en.wiktionary.org/wiki/mergeable) but I'm not a native speaker so I might be wrong.


jaredcwhite:
@alexanderadam Yeah, I think you're right! Can you add an `alias_method` so the typo version is still around and we don't introduce a breaking change?

alexanderadam:
@jaredcwhite I updated the PR

jaredcwhite:
Looks great! ✌️

PR #640, opened 2022-10-15
fix Lint/ConstantDefinitionInBlock and Style/StructInheritance

lxxxvi:

## Summary

This PR addresses 2 Rubocop violations that existed in `bridgetown-core/test/test_filters.rb`

## Context

* Moving Struct definitions outside a block resolves the `Lint/ConstantDefinitionInBlock` violation.
* At the same time `Style/StructInheritance` could be resolved by using `Foo = Struct.new {}` instead of `class Foo < Struct.new; end`

### DIFF
--- a/bridgetown-core/test/test_filters.rb
+++ b/bridgetown-core/test/test_filters.rb
@@ -31,6 +31,23 @@ class TestFilters < BridgetownUnitTest
     def select; end
   end
 
+  M = Struct.new(:message) do
+    def to_liquid
+      [message]
+    end
+  end
+
+  T = Struct.new(:name) do
+    def to_liquid
+      {
+        "name" => name,
+        :v     => 1,
+        :thing => M.new(kay: "jewelers"),
+        :stuff => true,
+      }
+    end
+  end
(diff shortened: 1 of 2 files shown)


bridgetown-core/test/test_filters.rb
```diff
@@ -31,6 +31,23 @@ class SelectDummy
     def select; end
   end
 
+  M = Struct.new(:message) do
```
andrewmcodes, line 34: I actually prefer this struct style personally 👏

andrewmcodes (approved):
Thanks so much @lxxxvi, I love everything about this 😄 

Removing overrides is a win in my book and I actually think this makes the code more readable.

bridgetown-core/test/test_filters.rb
KonnorRogers: +1! The inheriting of an initialized strict feels...weird, and I'm all for keeping Ruby weird...

jaredcwhite:
Thanks @lxxxvi!

PR #644, opened 2022-10-25
Resolve Rubocop offenses

vvveebs:
Reviewers - let me know if you want these squashed down to a single commit.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/lit.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/lit.rb
@@ -38,7 +38,7 @@ gsub_file "esbuild.config.js", %r{const esbuildOptions = {}\n} do |_match|
       // Read https://www.bridgetownrb.com/docs/components/lit#sidecar-css-files for documentation.
       /*
       postCssPluginConfig: {
-        filter: /(?:index|\.global)\.css$/,
+        filter: /(?:index|\\.global)\\.css$/,
       },
       */
     }
@@ -59,7 +59,7 @@ unless found_match
       // Read https://www.bridgetownrb.com/docs/components/lit#sidecar-css-files for documentation.
       /*
       postCssPluginConfig: {
-        filter: /(?:index|\.global)\.css$/,
+        filter: /(?:index|\\.global)\\.css$/,
       },
       */
     JS
--- a/bridgetown-core/lib/bridgetown-core/configurations/stimulus.rb
+++ b/bridgetown-core/lib/bridgetown-core/configurations/stimulus.rb
@@ -36,7 +36,7 @@ append_to_file(File.join(javascript_dir, "index.js")) do
       Object.entries(controllers).forEach(([filename, controller]) => {
         if (filename.includes("_controller.") || filename.includes("-controller.")) {
           const identifier = filename.replace("./controllers/", "")
-            .replace(/[_-]controller\..*$/, "")
+            .replace(/[_-]controller\\..*$/, "")
             .replace("_", "-")
             .replace("/", "--")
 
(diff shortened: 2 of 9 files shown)


bridgetown-core/lib/bridgetown-core/configurations/lit.rb
```diff
@@ -59,7 +59,7 @@
       // Read https://www.bridgetownrb.com/docs/components/lit#sidecar-css-files for documentation.
       /*
       postCssPluginConfig: {
-        filter: /(?:index|\.global)\.css$/,
+        filter: /(?:index|\\.global)\\.css$/,
```
andrewmcodes, line 62: So I noticed that this actually changes the output, which I don't think we want.

```diff
- filter: /(?:index|.global).css$/,
+ filter: /(?:index|\.global)\.css$/,
```

I'm no regex expert but given this, I don't think we want this change or the one on line 41. Wdyt?

bridgetown-core/lib/bridgetown-core/configurations/stimulus.rb
```diff
@@ -36,7 +36,7 @@
       Object.entries(controllers).forEach(([filename, controller]) => {
         if (filename.includes("_controller.") || filename.includes("-controller.")) {
           const identifier = filename.replace("./controllers/", "")
-            .replace(/[_\-]controller\..*$/, "")
+            .replace(/[_-]controller\\..*$/, "")
```
andrewmcodes, line 39: Similarly, these changes appear to change the output of this configuration as well.

```diff
- .replace(/[_-]controller..*$/, "")
+  .replace(/[_-]controller\..*$/, "")
```

Again, I don't think we want to include these.

andrewmcodes (changes requested):
Thanks for making this PR! I like the rule change to eliminate unnecessary escapes but it appears that some of these changes are unsafe and have changed the output of these bundled configurations.

Unless @jaredcwhite objects, I would disable this rule for the `bridgetown-core/lib/bridgetown-core/configurations/**/*.rb` folder. 

I also personally like to squash these but it can be squashed on merge so I'll leave that up to you.

jaredcwhite:
@andrewmcodes @vvveebs I agree, I would like to avoid changing anything that ends up as output in user projects at the present moment.

vvveebs:
Thanks both.

I remember having some fun with the config files and their output when I made this :)

I believe there are a few little edge case bugs which I was attempting to fix here rather than add an overarching Rubocop ignore rule but I probably didn't do a great job of explaining where I was coming from.

If I don't hear from folks in the coming days, I'll make a dedicated issue and add an ignore rule so we can move on :)

Essentially I think we are ending up creating a couple of regexes which contain the "any character" character class (`.`)  rather than the period character (`\.`) and I think we want the latter.

It's best to give an example in JS. Taking the example of the post CSS filter regex:

```
> const mainBranchLitRegex = /(?:index|.global)\.css$/  // this is a regex extracted from esbuild.config.js if you run `bin/bridgetown config lit` on main branch

> "somethingglobal.css".match(mainBranchLitRegex)  // the [docs](https://www.bridgetownrb.com/docs/components/lit#sidecar-css-files) say the expected suffix is '.global.css' but this string matches despite not having that suffix
```

A similar pattern seems to be in the stimulus config which modifies controller filenames. I doubt anyone has been stung by these edge cases but I figured best to try to fix them (with an additional escape `\`) if it means we can avoid a Rubocop ignore rule in case that ends up hiding something lurking for us in the future.

andrewmcodes:
@vvveebs well TIL 😅

andrewmcodes (approved):
If these are correct as @jaredcwhite  mentioned than this is good to go from my perspective! Thanks @vvveebs!

jaredcwhite:
Thanks for getting this over the edge @vvveebs. 🙏🏻  There might be a couple more things to massage here or there but probably makes sense to do that as new PRs off of main. Merging!

PR #645, opened 2022-10-28
Fix typo in confirmation prompt

brandoncc:
This is a 🔦 documentation change. (not really, but this is the closest option)

## Summary

Fixes a typo in a command line prompt that is shown while configuring tailwind.


jaredcwhite:
Thanks @brandoncc!

PR #651, opened 2022-11-16
Specify locales in "multi locale" mode

rickychilcott:
This is a 🙋 feature or enhancement with a 🔦 documentation change.

## Summary

Extending the ability of the "multi locale" mode to specify a limited set of locales instead of outputting all locales (i.e. `site.config.available_locales`)

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/collection.rb
+++ b/bridgetown-core/lib/bridgetown-core/collection.rb
@@ -279,10 +279,15 @@ module Bridgetown
       )
 
       if model_is_multi_locale?(model, model_relative_path)
-        site.config.available_locales.each do |locale|
+        # If the model specifies a locales key, use that to determine the
+        # the locale of each resource, otherwise fall back to `site.config.available_locales`
+        locales = model.locales || site.config.available_locales
+
+        locales.map do |locale|
           model.locale = locale
           add_resource_from_model model
         end
+
         return
       end
 
--- a/bridgetown-core/lib/bridgetown-core/collection.rb
+++ b/bridgetown-core/lib/bridgetown-core/collection.rb
@@ -271,6 +271,7 @@ module Bridgetown
     # Read in resource from repo path
     # @param full_path [String]
     # @param manifest [Bridgetown::Configuration::SourceManifest]
+    # @return [void]
     def read_resource(full_path, manifest: nil)
       model_relative_path = relative_model_path_for(full_path, manifest: manifest)
       model = Bridgetown::Model::Base.find(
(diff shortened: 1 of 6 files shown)


bridgetown-core/lib/bridgetown-core/collection.rb
```diff
@@ -279,10 +279,15 @@ def read_resource(full_path, manifest: nil)
...
+        # If the model specifies a locales key, use that to determine the
+        # the locale of each resource, otherwise fall back to `site.config.available_locales`
+        locales = model.locales || site.config.available_locales
+
+        locales.map do |locale|
           model.locale = locale
           add_resource_from_model model
         end
+
         return
```
rickychilcott, line 291: Unrelated to this PR @jaredcwhite - the return value of this method is inconsistent. When adding a multi-locale model, it returns nil, but when it doesn't have multi, then it returns the output of `add_resource_from_model model`.

This may not be a bug, and I tried to trace it back to see what the effects were, but I couldn't quite follow. Just highlighting this since I noticed it.

I'm realizing now that I left the `locales.map` when I meant to make it `locales.each`, but... maybe this should change to:

```ruby
return locales.map do |locale|
  model.locale = locale
  add_resource_from_model model
end
```

rickychilcott:
I tried to get a minimial shim of this into Bridgetown 1.1 and I got stuck because I'm not actually using this via files. Instead we are using the `add_resource` api from bridgetown-core... So I'm a bit stuck as to how best to solve my issues without also changing `add_resource` which doesn't seem right. Any maintainer willing to pair on this one?

bridgetown-core/lib/bridgetown-core/collection.rb
jaredcwhite: @rickychilcott I think the return value of this method can be void, so fine with refactoring to do that and adding a `@return [void]` to the comment doc.

jaredcwhite:
@rickychilcott Hmm, I don't think the "multi" mode really makes sense for the manual API to generate resources. Just call `add_resource` for each locale right? (So if you had 10 pages, and 3 locales each, you'd be calling `add_resource` a total of 30 times)

rickychilcott:
> @rickychilcott Hmm, I don't think the "multi" mode really makes sense for the manual API to generate resources. Just call `add_resource` for each locale right? (So if you had 10 pages, and 3 locales each, you'd be calling `add_resource` a total of 30 times)

We're doing some weird stuff -- fun but non-standard. But I totally agree after playing with it some more this evening.

bridgetown-core/lib/bridgetown-core/collection.rb
rickychilcott: 👍

jaredcwhite:
Thanks again @rickychilcott, this looks great!

PR #652, opened 2022-11-18
fix: update broken initializers documentation link

andrewmcodes:
This is a 🔦 documentation(ish) change.

## Summary

Replace documentation link in the initializers site template file that points to the wrong URL.

## Context

The site template initializer file points to `/documentation/` instead of `/docs/` resulting in a page not found error when clicking the link. 

As some added context, I couldn't actually find the initializers documentation for a moment, maybe it would be helpful to link to it in some predominate places/on the sidemenu?


andrewmcodes:
Looks like the tests are also failing on main

PR #654, opened 2022-11-23
feat: Localize file-based dynamic routes

tommasongr:
## Summary

Adds localized routes under the `_routes` folder. This is done by creating all the possible slugs on the fly based on `site.config.available_locales`.

Every route will set `I18n.locale`  and `r.params[:locale]` to the corresponding value extracted from the url slug.

A possible missing piece might be a dedicated config for opting-in/out from the behavior. However, I think that forcing the locale on the file frontmatter could be a good enough alternative and possibly a more granular one.

### DIFF
--- a/bridgetown-routes/lib/bridgetown-routes/manifest.rb
+++ b/bridgetown-routes/lib/bridgetown-routes/manifest.rb
@@ -32,7 +32,19 @@ module Bridgetown
                 ":#{Regexp.last_match(1)}"
               end
 
-              [file, file_slug, segment_keys]
+              # generate localized file slugs
+              localized_file_slugs = []
+              site.config.available_locales.each do |locale|
+                localized_file_slugs.push(
+                  if locale == site.config.default_locale && !site.config.prefix_default_locale
+                    file_slug
+                  else
+                    "#{locale}/#{file_slug}"
+                  end
+                )
+              end
+
+              [file, localized_file_slugs, segment_keys]
             end
 
             new_manifest += sort_routes!(routes)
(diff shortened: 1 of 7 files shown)


bridgetown-routes/lib/bridgetown-routes/manifest.rb
```diff
@@ -32,7 +32,19 @@ def generate_manifest(site) # rubocop:todo Metrics/AbcSize, Metrics/CyclomaticCo
                 ":#{Regexp.last_match(1)}"
               end
 
-              [file, file_slug, segment_keys]
+              # generate localized file slugs
+              localized_file_slugs = []
+              site.config.available_locales.each do |locale|
```
jaredcwhite, line 37: We could switch this to using map for simpler code, aka `localized_file_slugs = site.config.available_locales.map do |locale|`

jaredcwhite (approved):
@tommasongr Very nicely done. I had one little code nitpick, but otherwise LGTM. 👍

bridgetown-routes/lib/bridgetown-routes/manifest.rb
tommasongr: Thanks! I'm glad you liked the implementation. This change looks cleaner to me. I'll fix it right away!

jaredcwhite:
@tommasongr Looks like some Rubocop checks are failing. I'd be fine with adding some `# rubocop:todo` comments in if need be, but want to get your eyes on it first.

tommasongr:
@jaredcwhite Thanks for reporting it. Yeah, I was aware of that. I fixed all the "real" offenses, but I wasn't sure how to treat the Metrics one. Anyway I'll comment/fix all the remaining offenses and push a final commit. Give me a sec.

tommasongr:
@jaredcwhite I resolved three offenses in `manifest.rb` and `manifest_router.rb` by extracting some code to external methods and I marked as `rubocop:todo` an offense in `test_routes.rb`. I hope the new structure is fine with you. Let me know.

jaredcwhite (approved):
LGTM!

PR #658, opened 2022-11-25
Support for gems.rb

svoop:

(Note: Rubocop errors prevent the tests to run, I've tested manually for now and wait for #644 to be merged. However, I create the PR now to get your comments on the changes already.)

## Summary

Bundler added `gems.rb`/`gems.locked` as alternatives to `Gemfile`/`Gemfile.lock` some years ago, mainly to have proper syntax coloring in editors.

Bridgetown has a few places where `Gemfile` is hardcoded, this PR adds support for `gems.rb`. However, in order to assure downwards compatibility, the fallback `Gemfile` is maintained.

Thanks for reviewing and commenting!

### DIFF
--- a/bridgetown-core/lib/bridgetown-core.rb
+++ b/bridgetown-core/lib/bridgetown-core.rb
@@ -16,6 +16,7 @@ end
 
 # rubygems
 require "rubygems"
+require "bundler/shared_helpers"
 
 # stdlib
 require "find"
--- a/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/concerns/actions.rb
@@ -56,8 +56,9 @@ module Bridgetown
         options = +""
         options += " -v \"#{version}\"" if version
         options += " -g #{group}" if group
+        # in_bundle? returns the path to the gemfile
         run "bundle add #{gemname}#{options}",
-            env: { "BUNDLE_GEMFILE" => File.join(destination_root, "Gemfile") }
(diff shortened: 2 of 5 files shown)


jaredcwhite:
I'll take a look at this shortly!

jaredcwhite:
Thanks @svoop. This is starting to feel like we're rewriting Bundler logic, so I poked around in the Bundler gem and it looks like indeed there are some helpful methods in the `Bundler::SharedHelpers` module. For example:

```ruby
require "bundler/shared_helpers"

Bundler::SharedHelpers.in_bundle?
```

So maybe we could refactor this just to call out to relevant Bundler methods, so we don't need to maintain our own logic to determine/find gem file paths. See: https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/shared_helpers.rb

svoop:
@jaredcwhite This makes a lot of sense! Want me to have a go or do you rather want to refactor yourself?

jaredcwhite:
@svoop Feel free to jump right into it, but I'm happy to pair or offer any assistance!

svoop:
@jaredcwhite Sorry for dropping the ball here, I've changed the implementation to use `in_bundle?`. (Kinda weird it returns `nil` or the path instead of `true/false`, Rubocop should throw a tantrum. :smile:) It's all green, but I'm not sure whether it's fully covered by tests.

jaredcwhite:
@svoop I think we're fine for now. Seeing some more Rubocop failures but that seems to be unrelated to your changes so I'll go ahead and merge and then fix. Thanks!

svoop:
Cool, thanks @jaredcwhite

PR #663, opened 2022-12-02
feat: Add new Notes command to the CLI

tommasongr:
## Summary

Adds a new `notes` command to the Bridgetown CLI. It is basically a porting of the Rails `notes` command with some Bridgetown specific configurations.

### Example

Like in Rails, the command accepts an `annotations` option to configure the notes parsing:

CleanShot 2022-12-02 at 13 47 10@2x

This is an example output with no `annotations` option:

CleanShot 2022-12-02 at 13 46 33@2x

### Why

The `notes` command in Rails is useful by itself. However it is even more powerful when used by other tools. For example my Nova's extension [Ruby on Rails](https://github.com/tommasongr/nova-rails) uses it under the hood to display annotations in the sidebar:

CleanShot 2022-12-02 at 13 26 52@2x

I would love to support the same functionality for Bridgetown projects too.

### Bridgetown specific configurations

The implementation differs from the Rails one mainly in two ways:

- Parsable directories: `src, server, plugins, frontend, config, db, test`
- Parsable files: `liquid, serb, html, md, json` in addition to `builder, rb, rake, ruby, yml, yaml, css, js, erb`

The following regex `/{%\s*#\s*(#{tag}):?\s*(.*?)\s*%}/` is used for parsing `liquid, serb` files.
The following regex `/[<|{]%\s*#\s*(#{tag}):?\s*(.*?)\s*%[}|>]/` is used for parsing `html, md, json` files.

If there are other file types we want to support, please let me know.

## TODO

Proper testing is currently missing. Honestly I'm not really sure how to do it... Can someone give me a hand? 🙏


jaredcwhite:
This seems like a great feature @tommasongr! I'm exited to start using it myself.

Perhaps for testing, there could be some of these notes added to files in the `test/source` or `test/resources` folders and then validate some of the command output in a new test? You can look at `test/test_doctor_command.rb` as an example of testing a command.

tommasongr:
Thanks @jaredcwhite, I'm excited too! As soon as this is going to be merged I'll publish a new Nova's extension just for Bridgetown.

Right now we already provide a task with the main three console commands in [Ruby on Rails](https://github.com/tommasongr/nova-rails), but I think that an extension with dedicated features could provide even more value (like the notes sidebar).

CleanShot 2022-12-05 at 22 08 11@2x

If you have any further ideas on how to improve the BT dev experience in an IDE, I'd be really interested to hear about them 😆

Also thanks for the suggestions about testing. I'll do some experiment and let you know.

jaredcwhite:
@tommasongr I'd be fine with merging this as-is if that's OK with you. Was there anything else you wanted to touch on here?

jaredcwhite:
Looks like we'd need to clean this up for linting to pass.

PR #674, opened 2022-12-18
Fix the readability of namespaces in dark mode

michaelherold:

## Summary

Without this, namespaces end up the same color as the background so they are unreadable. This change makes namespaces the same color as classes and constants.

```ruby
module Foo
    #  ^------ this is a namespace
  class Bar
    #   ^----- and it will be the same color as this
  end
end
```

## Context

None.


vvveebs:
Thanks @michaelherold

This looks good 👍

PR #677, opened 2022-12-21
Add some DX improvements to `StaticFile`

jaredcwhite:
Extracted from my blog and cleaned up.


vvveebs:
Is it worth adding a couple of entries for these in the table in the [static file docs](https://www.bridgetownrb.com/docs/static-files)?

jaredcwhite:
@vvveebs Good point…that table is sort of a legacy holdover that's about Liquid specifically, but I went ahead and added the `date` alias.

Issue #684, opened 2022-12-27
URL does not change to localhost in development when using initializers


michaelherold:
**Bridgetown Version**: bridgetown 1.2.0.beta4 "Bonny Slope"

**To Reproduce**

See [this repository](https://github.com/michaelherold/bt_initializer_bug) for a reproduction.

Manual steps:

1. Set `url: "https://example.com"` in `bridgetown.config.yml`
2. Start the development server with `bin/bt dev`
3. Use `page.relative_url | absolute_url` in a page and see that it's `http://localhost:4000`
4. Stop the development server and modify `config/initializers.rb` with `url "https://example.com"`
5. Restart the development server with `bin/bt dev`
6. Use `page.relative_url | absolute_url` in a page and see that it's now `https://example.com`

**Current behavior**

When using `config/initializers.rb` to set the `url` property for the site, it does not get overridden in development to `http://localhost:4000`.

**Expected behavior**

The `url` property should be overridden locally just like when you set `url` via `bridgetown.config.yml`.

**Computing environment (please complete the following information):**

 - OS: macOS 12.3.1
 - Browser:] Firefox
 - Browser Version: 102.1
 - Ruby Version: 3.1.2

**Additional context**

This is a useful feature that seems like it should work when set via an initializer too. Perhaps this is intended behavior, but it seems preferable to be able to get rid of `bridgetown.config.yml` entirely and configure your site from `config/initializers.rb` without regressions.

jaredcwhite:
Fixed with #686

PR #686, opened 2022-12-28
684-restore-localhost-url-with-initializers

jaredcwhite:
Fixes issue #684

Also removes the missing configuration file warning if an initializers-based config is available

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configuration.rb
+++ b/bridgetown-core/lib/bridgetown-core/configuration.rb
@@ -130,7 +130,7 @@ module Bridgetown
       end
     end
 
-    def run_initializers!(context:)
+    def run_initializers!(context:) # rubocop:todo Metrics/AbcSize, Metrics/CyclomaticComplexity
       initializers_file = File.join(root_dir, "config", "initializers.rb")
       return unless File.file?(initializers_file)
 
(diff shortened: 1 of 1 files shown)


bridgetown-core/lib/bridgetown-core/configuration.rb
```diff
@@ -144,9 +144,11 @@ def run_initializers!(context:)
       Bridgetown.logger.debug "Initializing:", "Running initializers with `#{context}' context in:"
       Bridgetown.logger.debug "", initializers_file
       self.init_params = {}
+      cached_url = url&.include?("//localhost") ? url : nil
```
michaelherold, line 147: **question:** How else other than running `bin/bt dev` can the value be set at this point? I'm unclear why we need to grab this `cached_url` because of my lack of understanding here.

Do we need to grab it because the port is overrideable? Because it can be set in other ways?

My hypothesis is that perhaps a `bridgetown.config.yml` has already been processed and we need to account for a value in there. Am I on the right track?

jaredcwhite: @michaelherold it's overridden by the `start` command when in development environment before any other configuration has been loaded, even if there's a `url` key in the YAML config. But then when the initializers code is executed, setting `url` was overriding the `localhost` overwrite 😅. We want to undo that if necessary. There may be a more elegant way to refactor this in the near future…

PR #700, opened 2023-01-24
fix: load `initializers.rb` instead of require for working tests

jaredcwhite:
## Summary

This changes the code which executes Bridgetown's Ruby configuration DSL so initializer files in `config` are loaded via `load` rather than `require`

## Context

In testing scenarios (such as plugin gems) when `run_initializers!` may be called many times across a series of tests, the available initializers were blank after the first run. That's because the file to set up those initializers wasn't getting executed by Ruby. We need to `load` rather than `require` those files.

I'm pretty sure this isn't a breaking change, but we'll want to keep an eye on any odd behavior (possibly due to some mistake in attempting to run initializers more than once on the same site config)


ayushn21 (approved):
Awesome 👏  ... all looks good to me ... tested on the `bridgetown-sitemap` plugin off this branch and it's green

PR #702, opened 2023-01-24
Add method to reset config for test suites

jaredcwhite:
This adds a `Bridgetown.reset_configuration!` method which test suites can call before creating new configuration objects prior to site instantiation. This is required so that previously-defined initializers (other than for the site itself i.e. loaded via `config/*`) / source manifests are available once again to a new site config.


ayushn21 (approved):
👏

PR #706, opened 2023-02-01
Make the Roda app class lazy in applying plugins

michaelherold:

## Summary

Currently, when the `bridgetown-core/rack/roda.rb` is required — which happens as soon as you require `bridgetown-core/rack/boot.rb` — it starts to apply plugins to the base class. This leads to tricky behavior when writing tests around Bridgetown Roda apps; you have to set up the configuration ahead of time and juggle configuration when you want to create several apps with different configurations.

This change reorders the plugin initialization to happen once you inherit `Bridgetown::Rack::Roda` instead of at load time. This has two distinct benefits:

1. You can now require the file without having done the configuration setup dance
2. Roda apps configure themselves with the preflight configuration as it exists at the time they are defined, meaning you can have several apps with differing configurations

It also meshes better with my understanding of the lifecycle of a Roda app. Since Roda is intentionally designed to allow you to create multiple apps, having the base class for Bridgetown Roda app have a centralized configuration upon loading of the file feels askew with Roda's intent.

## Context

I wasn't sure how to add a test for this. I figured that if the test suite passed with this refactor, everything was okay with the change. If you have thoughts on adding a test, I'm open to doing so!

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/rack/roda.rb
+++ b/bridgetown-core/lib/bridgetown-core/rack/roda.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "roda/plugins/exception_page"
+
 unless Bridgetown::Current.preloaded_configuration
   raise "You must supply a preloaded configuration before loading Bridgetown's Roda superclass"
 end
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Thanks @michaelherold. Some really good points you make. It might take me a little bit to wrap my head around how this change might affect everything. There's a possibility, if we go down this path, that we eschew using a custom subclass at all and just subclass directly from Roda, and then use a mixin or base plugin to add in all the specific BT functionality. I look forward to tinkering with it a bit and working with you on improving the Roda integration.

jaredcwhite:
Sorry for the wait on this @michaelherold (and a few other of your PRs) — I hope to have a "PR review day" here soon for Bridgetown and will get to this. 🙏🏻

michaelherold:
No worries, Jared! I've seen all your posts recently; you're busy. 😄

The testing harness is one of the trickiest things I've found in working on Bridgetown plugins, but I have been able to fall back to writing integration tests based on files instead of memory-based fixtures. If this isn't a high priority, I understand!

jaredcwhite:
Closing in favor of #737 — if you have any feedback there @michaelherold let me know!

PR #712, opened 2023-02-04
Fix typo in initializers docs

fbuys:
A very simple typo fix.
But please let me know if I misunderstood the docs and this is not a typo after all.


jaredcwhite:
Technically, hmm…not necessarily a typo per se, but your description makes much more sense so let's go with that. Thanks!

PR #715, opened 2023-02-06
Clarify gem based plugin initializer instructions

fbuys:
## Summary

Updated a section of the gem based plugin's documentation.

## Context

I struggled to get my plugin to work.
It was unclear where I needed to add the plugin's initializer.

I referred to this plugin to figure out where to put the initializer.
https://github.com/ayushn21/bridgetown-sitemap/blob/main/lib/bridgetown-sitemap.rb

I also then saw an error: `undefined method `builder' for main:Object` which I fixed by updating that line to `config.builder`.

### DIFF
--- a/bridgetown-website/src/_docs/plugins.md
+++ b/bridgetown-website/src/_docs/plugins.md
@@ -108,12 +108,12 @@ module MyNiftyPlugin
   end
 end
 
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Thanks @fbuys for the corrections!

Issue #716, opened 2023-02-07
Configuration options not set by init in initializers.rb


fbuys:
**Bridgetown Version**: bridgetown 1.2.0 "Bonny Slope"      

**To Reproduce**

- Add config to initializers.rb
- Start app or Run tests
```ruby
# initializers.rb
  init :sample_plugin do
    Bridgetown.logger.info 'sample initializer...'
    my_setting "test-config-option"
  end
```

- Here is repo (branch: `initializers-issue`): https://github.com/fbuys/bridgetown-sample-plugin/tree/initializers-issue
- Clone repo and checkout `initializers-issue` branch
- `bundle install` and then run `rake test`

**Current behavior**

- Notice that config options are not passed to the plugin initializer
- But the init params are set on the config object

Inside the plugin initializer I am outputting like this:
```ruby
  pp config.to_h
  print "sample_plugin config: " 
  pp config.sample_plugin
```

```shell
...[rest of config object],
 "init_params"=>{"sample_plugin"=>{"my_setting"=>"test-config-option"}},
 "sample_plugin"=>{}}
sample_plugin config: {}
```

**Expected behavior**

I would like to access the config options set in the app initializer.

```ruby
  print "sample_plugin config: " 
  pp config.sample_plugin
```
should output:
```shell
sample_plugin config: { my_setting: "test-config-option" }
```

**Computing environment (please complete the following information):**
 - OS: [e.g. iOS] MacOS
 - Browser [e.g. chrome, safari] n/a
 - Browser Version [e.g. 22] n/a
 - Ruby Version (check with ruby -v) ruby 3.0.1p64 or ruby 3.1.2p20

jaredcwhite:
@fbuys I think you'd need to add `my_setting` as a keyword argument to your initializer for this to work: https://github.com/fbuys/bridgetown-sample-plugin/blob/47f37124f50c81b964c52ba8a981e0f398cda411/lib/sample_plugin.rb

There's probably somewhere in the documentation we should illustrate this more clearly. Also it's best to keep `my_setting` as an optional kwarg (aka default to nil or some value) unless you absolutely need it.

fbuys:
Hey I pushed a PR that might be helpful.

PR #717, opened 2023-02-07
Clarify adding keyword args to plugin initializers

fbuys:

## Summary

This PR adds some context to how we can use keyword arguments to pass values from an end user's initializers.rb file into our plugin's initializer.

## Context

Closes: https://github.com/bridgetownrb/bridgetown/issues/716


jaredcwhite:
Thanks @fbuys, this looks great. I think it would be ideal for the kwarg and the config key to match, would you be able to update that so it's just `api_key` rather than `nifty_api`?

fbuys:
Great idea, I updated the paragraph and example.

PR #722, opened 2023-02-26
Fix passing headers and add passing parameters to HTTP DSL

### LINKED ISSUE #630: feat: Params support for the `get` builder DSL
Strangely, you can't pass params to the `get` method in builders. That's…weird! I'd expect it to take something like a `params` kwarg. Also…there might be a bug with how the headers are getting passed into Faraday's own `get` method.

Seems like we'll want to update things here:
`bridgetown-builder/lib/bridgetown-builder/dsl/http.rb`

Reference: https://lostisland.github.io/faraday/usage/

michaelherold:
This is a 🐛 bug fix.

## Summary

This change includes two related items:

1. Fixes the handling of headers for HTTP DSL requests. They were improperly being sent as a `headers` parameter and now will be sent as actual headers.
2. Adds sending parameters to the URL via keyword arguments. This is valuable because constructing the query string parameters directly is tedious and prone to typos.

## Context

Closes #630

### DIFF
--- a/bridgetown-builder/lib/bridgetown-builder/dsl/http.rb
+++ b/bridgetown-builder/lib/bridgetown-builder/dsl/http.rb
@@ -10,7 +10,7 @@ module Bridgetown
       module HTTP
         def get(url, headers: {}, parse_json: true)
           body = begin
-            connection(parse_json: parse_json).get(url, headers: headers).body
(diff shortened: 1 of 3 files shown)


vvveebs (commented):
This looks good

PR #723, opened 2023-02-26
Make OutputBuffer more performant for 3.2+ YJIT

### LINKED ISSUE #599: Implement Ruby 3.2+ speedup for template output buffers
As our output buffer code is heavily based on the Rails implementation, we should look into updating our code to mirror the speed gains Rails just facilitated when run on upcoming Ruby versions. See: https://github.com/rails/rails/pull/45614

michaelherold:

## Summary

This change ports Jean Boussier's Rails change over to Bridgetown with a few tweaks.

1. Using `instance_of?` against the constant is faster than the alternative `self.class == other.class` construct
2. Linear, comparable logic for `<<`
3. An expanded set of delegated methods to include `#empty?`, `#encode`, `#lines`, `#reverse`, `#strip`, and `#valid_encoding?` because these are all called in the test suite.

In the current test suite, this is a count of how many times each method is called:

```
2794 empty?
  54 encode
  54 encoding
 381 lines
 127 reverse
 127 strip
  54 valid_encoding?
```

It's unclear to me which, if any, of these should be public API so I kept all of the delegated methods as well as the ones Jean picked for Rails.

## Context

Closes #599

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
+++ b/bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
@@ -3,18 +3,57 @@
 require "tilt/erubi"
 
 module Bridgetown
-  class OutputBuffer < ActiveSupport::SafeBuffer
-    def initialize(*)
-      super
-      encode!
+  class OutputBuffer
+    def initialize(buffer = "")
+      @buffer = String.new(buffer)
+      @buffer.encode!
     end
+
+    def initialize_copy(other)
+      @buffer = other.to_str
+    end
+
+    delegate(
+      :blank?,
+      :empty?,
+      :encode,
+      :encode!,
+      :encoding,
+      :force_encoding,
+      :length,
+      :lines,
+      :reverse,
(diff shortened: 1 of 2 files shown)


sandstrom:
Sorry for the skepticism, but is it worth maintaining this extra code, over using ActiveSupport::SafeBuffer directly?

Rails render HTML directly, but BT is a static site builder. Performance in the build step is less of an issue, generally speaking.

Would it be possible to import the `ActionView::OutputBuffer` class directly from `action_view` instead?

That way, we don't have to maintain this.

Just a thought though, feel free to disregard.

michaelherold:
That seems reasonable to me. This was me looking at the open issues list and finding one that I could do. I'm not sure why we don't currently use `ActionView::OutputBuffer` directly.

jaredcwhite:
It looks like this is a good fit for merging in along with #1040. Get on a newer Erubi paradigm _and_ reduce our AS dependency! Thanks for working on this @michaelherold.

PR #728, opened 2023-03-13
Avoid invisible code highlight for bold text in dark mode

neilvanbeinum:

## Summary

Avoids invisible bold text in dark mode.

Example docs page showing the issue (when in dark mode): https://www.bridgetownrb.com/docs/front-matter#the-power-of-ruby-in-front-matter

New styling example below:

Screenshot 2023-03-13 at 17 33 10

### DIFF
--- a/bridgetown-website/frontend/styles/global.css
+++ b/bridgetown-website/frontend/styles/global.css
@@ -27,6 +27,7 @@
   --color-syntax-beige: #8A786E;
   --color-syntax-charcoal: #67717E;
   --color-syntax-steel-teal: #568684;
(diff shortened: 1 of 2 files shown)


jaredcwhite:
Thanks @neilvanbeinum, good catch. Can we extract the `#4b83d4` out to a CSS variable so it's more DRY? (I noticed it used in more than one place)

neilvanbeinum:
Thanks @jaredcwhite - updated with the CSS variable [here](https://github.com/bridgetownrb/bridgetown/pull/728/commits/42dbf79187d7df5e655380ead48b4a6bda6576b0)

PR #733, opened 2023-03-22
Fix purgecss plugin crashing when using scss

goulvench:
When using SASS, the manifest key is `index.scss` instead of `index.css`, raising `undefined method 'split' for nil:NilClass`.
The crash happens at line 24 (`css_file = manifest["styles/index.css"].split("/").last`).

This is a 🐛 bug fix.

I haven't found tests for this plugin, let me know if there are that I can tweak to exercise this condition.

## Summary

When a project uses scss instead of plain css, the expected manifest key is slightly different, crashing the `purgecss` plugin to crash. By reading either `index.css` or `index.scss` the plugin handles both cases without erroring out.


jaredcwhite:
Looks good to me! Thanks @goulvench

PR #737, opened 2023-04-01
Relocate all Bridgetown Roda logic to new `bridgetown_stack` plugin

jaredcwhite:
This is an alternate take on the concepts brought forth by @michaelherold in https://github.com/bridgetownrb/bridgetown/pull/706

Rather than have `RodaApp` subclass from a Bridgetown-specific subclass of `Roda`, you can simply subclass from `Roda` like any standard Roda application, and then use the `bridgetown_stack`\* plugin to pull in all of Bridgetown's setup.

This provides the added benefits of future improvements where some of the ways in which the stack plugin sets up the app could be further configurable as plugin options.

\* Still thinking about the best name to call this. Also there's opportunity to consolidate some of the existing plugin(s) into this new one.


jaredcwhite:
@michaelherold I'm curious what your thoughts are on this PR!

ayushn21:
Can we just call the plugin `bridgetown`? I think a Roda plugin called `bridgetown` makes the most sense if we can do that ...

jaredcwhite:
@ayushn21 I believe I tried that and ran into some annoying namespace issues, but I'll see if I can come up with a workaround. I agree that would probably be ideal.

brandondrew:
`bridgetown_server` seems like the clearest possible name 👍

jaredcwhite:
LGTM!

michaelherold:
I **think** that this will make testing easier without using integration tests, but haven't had the creative energy available to test it. Good idea, Jared!

Once I am able, I will post an experience report on the new structure.

PR #739, opened 2023-04-08
Fix permalink file extensions, if basename contains "."

Skudo:
This is a 🐛 bug fix.

## Summary

When the filename of a permalink contains more than one dot, e. g. `foo.bar.html`, we strip its file extension before processing it further. The basename without extension now is `foo.bar`.

https://github.com/bridgetownrb/bridgetown/blob/715cf3a4460952c2f7d7a17f2bde6c2e3d36d6dd/bridgetown-core/lib/bridgetown-core/resource/permalink_processor.rb#L35

https://github.com/bridgetownrb/bridgetown/blob/715cf3a4460952c2f7d7a17f2bde6c2e3d36d6dd/bridgetown-core/lib/bridgetown-core/filters/url_filters.rb#L71-L75

When the final permalink is built with all of its placeholders filled, the original file extension is added back. However, when the file extension to be re-added is determined, a regular expression is used.

https://github.com/bridgetownrb/bridgetown/blob/715cf3a4460952c2f7d7a17f2bde6c2e3d36d6dd/bridgetown-core/lib/bridgetown-core/resource/permalink_processor.rb#L87-L88

For a filename with multiple dots like the aforementioned `foo.bar.html`, that regular expression considers `.bar.html` the file extension. Putting the basename and the extension together thus results in `foo.bar.bar.html` instead of `foo.bar.html`.

This PR fixes that issue by determining the file extension through `File.extname` instead of using the aforementioned regular expression. This feels like the most robust way; updating the regular expression to `%r{\.[^./]*$}` would work, as well, but since `Bridgetown::Filters::URLFilters` is responsible for removing the file extension, it should also be responsible for telling us what the file extension is. This is why I opted against the more obvious

```ruby
elsif permalink.match?(%r{\.[^./]*$})
  "/#{new_url}#{File.extname(permalink)}"
```

### DIFF
--- a/bridgetown-core/features/permalinks.feature
+++ b/bridgetown-core/features/permalinks.feature
@@ -92,6 +92,17 @@ Feature: Fancy permalinks
     And the output/custom/posts directory should exist
     And I should see "bla bla" in "output/custom/posts/some.html"
 
+  Scenario: Use per-post ending in .en.html
+    Given I have a _posts directory
+    And I have the following post:
+      | title     | date       | permalink                  | content |
+      | Some post | 2013-04-14 | /custom/posts/some.en.html | bla bla |
+    When I run bridgetown build
+    Then I should get a zero exit status
+    And the output directory should exist
+    And the output/custom/posts directory should exist
+    And I should see "bla bla" in "output/custom/posts/some.en.html"
+
   Scenario: Use pretty permalink schema with cased file name
     Given I have a _posts directory
     And I have an "_posts/2009-03-27-Pretty-Permalink-Schema.md" page that contains "Totally wordpress"
--- a/bridgetown-core/lib/bridgetown-core/filters/url_filters.rb
+++ b/bridgetown-core/lib/bridgetown-core/filters/url_filters.rb
@@ -74,6 +74,14 @@ module Bridgetown
         end.to_s
       end
 
+      # Returns the extension of a path/URL
+      #
+      # @param input [Object] value which responds to `to_s`
+      # @return [String]
+      def extname(input)
+        File.extname(input.to_s)
+      end
+
       private
 
       def compute_absolute_url(input)
--- a/bridgetown-core/lib/bridgetown-core/resource/permalink_processor.rb
+++ b/bridgetown-core/lib/bridgetown-core/resource/permalink_processor.rb
@@ -85,8 +85,8 @@ module Bridgetown
         if permalink.ends_with?(".*") || !%r{\.html?$}.match?(final_ext)
           "/#{new_url}#{final_ext}"
         # If permalink includes the file extension, add it back in to the URL
-        elsif permalink =~ %r{\.[^/]*$}
-          "/#{new_url}#{Regexp.last_match[0]}"
+        elsif permalink.match?(%r{\.[^/]*$})
+          "/#{new_url}#{Bridgetown::Filters::URLFilters.extname(permalink)}"
         # Ensure index-style URLs get output correctly
         elsif permalink.ends_with?("/")
           "/#{new_url}/".sub(%r{^/index/$}, "/")


brandondrew:
What's the purpose of stripping off the file extension in the first place, if it is just going to be added back later?

Skudo:
Good question that made me realise that my "fix" doesn't actually fix things in general, just for my specific use case.

A permalink with a placeholder as the basename and a suffix needs to temporarily remove the suffix to correctly evaluate the placeholder.

Example: `foo/bar/:title.html` for a post titled "An amazing title".

1. We remove the file extension. `foo/bar/:title`.
2. We split the permalink into its segments. `["foo", "bar", ":title"]`.
3. We evaluate placeholders. `["foo", "bar", "an-amazing-title"]`
4. We re-join the segments. `foo/bar/an-amazing-title`.
5. We re-append the removed file extension. `foo/bar/an-amazing-title.html`

With "multiple file extensions" like `foo/bar/:title.en.html`, the fix has to be "remove everything after the first dot in the basename, then append everything later". Otherwise, the segments in the above example would be `["foo", "bar", ":title.en"]` and we wouldn't be able to resolve placeholder correctly.

brandondrew:
> With "multiple file extensions" like `foo/bar/:title.en.html`, the fix has to be "remove everything after the first dot in the basename, then append everything later". Otherwise, the segments in the above example would be `["foo", "bar", ":title.en"]` and we wouldn't be able to resolve placeholder correctly.


Thanks for the explanation! 

Are there ever any cases where one of the file extensions must also be evaluated?  For example `foo/bar/:title.:locale.html`?  If that sort of scenario is (or might ever be) supported, then maybe the solution is 

1. split on `.` first,
2. split on `/` next (storing it in a different variable OR a different part of a hash)
3. evaluate each segment, and 
4. put it all back together

So `foo/bar/:title.:locale.html` might first get broken up into `["foo/bar/:title", ":locale", "html"]`.  
Then the first element might get broken down into `["foo", "bar", ":title"]`.
If there's value in having it all in one hash, it could be (for example) a hash named `segments`, containing:
```
  path: ["foo", "bar", ":title"],
  extension: [":locale", "html"]
```
By knowing which segments belong to the path and which to the extensions, they can easily be put back together properly.

Please let me know if I'm overlooking something.

brandondrew:
(After all, I'm assuming that if you have `en` in the extensions, you might also have `nl`, `fr`, `vi`, `de`, `es`, and many many others.)

jaredcwhite:
Thanks for looking into this. Let's leave the "multiple placeholders per segment" question to a different issue/conversation. That's currently unsupported, which is actually intentional, and if we were going to reverse that it would require a larger rethink.

For the original case of "why is my bespoke permalink resulting in duplicated extensions in the output file?", let's treat that as a bug and fix in the simplest way possible. My preferred fix would be this for lines 88-89 of the permalink processor:

```ruby
        elsif (permalink_ext = File.extname(permalink)).empty?.!
          "/#{new_url}#{permalink_ext}"
```

I don't think we need to touch any of the Liquid filters at all.

Skudo:
Disregarding multiple placeholders per segment, the suggested fix

```
        elsif (permalink_ext = File.extname(permalink)).empty?.!
          "/#{new_url}#{permalink_ext}"
```

only fixes permalinks like `/constant.foo.html`. If basename contains a placeholder _and_ has multiple extensions, e. g. `/:title.foo.html`, the placeholder segment will be `:title.foo`, which then won't be evaluated correctly, will it?

With that said, I'm fine with the suggested minimal fix, especially as it fixes my concrete issue. However, I do think there is value in having `URLFilters.extname` to determine the file extension though, as `URLFilters` also is the entity that decides what a file extension is (by removing it in `.strip_extname`). Do you think that's enough of a reason to introduce `URLFilters.extname`?

jaredcwhite:
@Skudo Yeah, I don't see `:title.foo.html` as being supported at this time. Generally either `/` or `.*` is the preferred permalink suffix.

Let me think about the URLFilters question…at heart those methods are all there for the benefit of Liquid templates, and it's probably a flaw/legacy design that we'd rely on that for any internal processing. And I hesitate to add something like `extname` as a Liquid filter as then the question becomes why don't we offer a whole lot of other native Ruby methods for all sorts of string processing as filters? (For context, we're switching away from Liquid as a default template option in the next major release.)

jaredcwhite:
I apologize this was left hanging for so long. I'm just going to go ahead and close at this point, I don't feel confident to understand why these exact changes were made or how better to do it. Generally speaking, I don't see any useful examples out in the wild of permalinks with multiple extensions, so I'm pretty comfortable saying it's just not supported by Bridgetown.

PR #759, opened 2023-05-08
Add I18n localization shortcut "l" alongside "t"

### LINKED ISSUE #753: feat: Add date localization helper
## Summary

Expose the `l` helper provided by `I18n`, besides the existing `t`, [like Rails does](https://guides.rubyonrails.org/i18n.html#the-public-i18n-api). 

## Motivation

It's easy to translate UI strings, but dates remain in English. One can use `I18n.l`, but since Bridgetown strives to make it easy for Rails people to feel at home, one would expect to have both `t` and `l` helpers.

In liquid
```smarty
{% l resource.data.date, date_with_weekday %} <!-- tag style -->
{{ resource.data.date | l date_with_weekday }} <!-- filter style -->
```

In erb
```erb
<%= l Time.now, format: :date_with_weekday %>
```

## Next step

Create a `bridgetownrb/bridgetown-i18n` gem mirroring [`rails-i18n`](https://github.com/svenfuchs/rails-i18n/), commented out by default in Gemfile. If added to the bundle, translations not provided in `src/_locales` would fallback to `bridgetown-i18n`.

goulvench:

Notes:
- I haven't found tests for `I18n.t` that I could duplicate, but maybe I haven't looked in the right places. I'm not familiar with minitest.
- I've adjusted the documentation.
- The ci/build script cannot run on my machine, rbenv refuses to compile ruby 2.7.2. 
- While browsing the code, I sumbled upon `date_to_string` and `date_to_long_string`, which could be rewritten using `I18n`. 

## Summary

Add `l` filter/tag/shortcut to `I18n.l`.

## Context

  Closes #753

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/filters/localization_filters.rb
+++ b/bridgetown-core/lib/bridgetown-core/filters/localization_filters.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Bridgetown
+  module Filters
+    module LocalizationFilters
+      def l(input)
+        I18n.l(input.to_s)
+      end
+    end
+  end
+end
--- a/bridgetown-core/lib/bridgetown-core/helpers.rb
+++ b/bridgetown-core/lib/bridgetown-core/helpers.rb
@@ -135,6 +135,14 @@ module Bridgetown
         I18n.send :t, *args, **kwargs
       end
 
+      # Forward all arguments to I18n.l method
+      #
+      # @return [String] the localized string
+      # @see I18n
+      def l(*args, **kwargs)
+        I18n.send :l, *args, **kwargs
+      end
+
       # For template contexts where ActiveSupport's output safety is loaded, we
       # can ensure a string has been marked safe
       #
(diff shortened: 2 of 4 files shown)


brandondrew:
@goulvench I am guessing that the issue with Ruby 2.7.2 is that you're on a Mac with Apple Silicon (M1 or M2) and that's before Ruby was released with support for Apple Silicon.  (I know I was unable to run 2.6.6 for that reason, and I was able to run 2.7.7, but I don't know exactly where the cutoff was.)

@jaredcwhite did you want to keep an old version of Ruby running the CI just to establish that such an old version will work?  

Would it be worth considering making 3.x a minimum for Bridgetown 1.3.0?  (Or at least 2.7.8?)

goulvench:
@brandondrew It's not an Apple silicon issue, I used to run ruby 2.5 a couple years ago but scrapped it since I no longer had any project using it. The issue is a null pointer crash when compiling `md5.c`, but since ruby 2.7.2 is past EOL and CI runs fine I'm gonna leave it at that :-)

Issue #734, opened 2023-03-24
Yarn PnP results in esbuild errors, Yarn classic not an issue


ivalentine:
**Bridgetown Version**: 1.2.0

**To Reproduce**

Create a new Bridgetown site, then run `bundle exec bridgetown start`.

**Current behavior**

When creating and starting a new site, I am encountering esbuild errors. The server process starts and I am able to view the site, but the CSS is broken. Here is the stack trace:
```
[Frontend] yarn run esbuild-dev
[Frontend] esbuild: frontend bundling started...
[Frontend] ✘ [ERROR] Could not resolve "index.css"
[Frontend]
[Frontend]     frontend/javascript/index.js:1:7:
[Frontend]       1 │ import "index.css"
[Frontend]         ╵        ~~~~~~~~~~~
[Frontend]
[Frontend]   The Yarn Plug'n'Play manifest forbids importing "index.css" here because it's not listed as a dependency of this package:
[Frontend]
[Frontend]     .pnp.cjs:962:33:
[Frontend]       962 │           "packageDependencies": [\
[Frontend]           ╵                                  ~~
[Frontend]
[Frontend]   You can mark the path "index.css" as external to exclude it from the bundle, which will remove this error.
[Frontend]
[Frontend] ✘ [ERROR] Could not resolve "syntax-highlighting.css"
[Frontend]
[Frontend]     frontend/javascript/index.js:2:7:
[Frontend]       2 │ import "syntax-highlighting.css"
[Frontend]         ╵        ~~~~~~~~~~~~~~~~~~~~~~~~~
[Frontend]
[Frontend]   The Yarn Plug'n'Play manifest forbids importing "syntax-highlighting.css" here because it's not listed as a dependency of this package:
[Frontend]
[Frontend]     .pnp.cjs:962:33:
[Frontend]       962 │           "packageDependencies": [\
[Frontend]           ╵                                  ~~
[Frontend]
[Frontend]   You can mark the path "syntax-highlighting.css" as external to exclude it from the bundle, which will remove this error.
[Frontend]
[Frontend] esbuild: build process error, cannot write manifest
[Frontend]   Exception raised: RuntimeError
[Frontend] Command failed with status (1): [yarn run esbuild-dev...]
[...]

PR #762, opened 2023-05-21
Update to esbuild 0.17

### LINKED ISSUE #697: Upgrade Esbuild to v0.17
**Bridgetown Version**: 1.2.0 beta5

Bridgetown is currently using Esbuild 0.16.x, it'd be nice to upgrade to v0.17. It has some breaking changes (eg. the `--watch`): https://github.com/evanw/esbuild/releases/tag/v0.17.0

jaredcwhite:
## Summary

In the process of upgrading esbuild to the latest version (along with PostCSS and other dependencies), I encountered some super weird issues with the watch process and CSS imports (aka index.css). Sometimes the files wouldn't rebuild at all when edited. Other times it would happen repeatedly.

I eventually tracked it down to some change back in 0.16.8+ around dependency resolution. There was a later change involving the `nodePaths` option, which seems not to have fully fixed this issue I saw.

I ended up going down a very different path. Rather than try to use esbuild's `nodePaths` config option in order to get nice imports like `import "index.css"` in `index.js`, I went with a more "standard" sort of path alias setup. esbuild can read in `jsconfig.json` and evaluate path aliases, which means you can add literally any alias you want. `import "foo/bar/file.js"` could resolve to "whatever/you/want/file.js" simply by adding aliases to `jsconfig.json`. Pretty sweet.

So…with this update we now have clearly defined path aliases for common frontend build paths. They are:

```
$styles: frontend/styles
$javascript: frontend/javascript
$components: src/_components
```

Thus in `index.js`, it should now have `import "$styles/index.css"`. The glob import can now have `import "$components/**/…"`, etc. And site devs can add their own additional path aliases if they want to! 🎉 

This also feels more future-compatible with other package managers and so forth.

> **Warning**
> This is a breaking change for existing Bridgetown sites, but **only** if they run the `esbuild update` command. And that command will point them to the need to update their CSS imports along with a link to the docs, so I can't imagine it's too big of an ask.

## Context

Resolves #697 and possibly #734 (not yet tested).

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
+++ b/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
@@ -307,5 +307,6 @@ module.exports = async (outputFolder, esbuildOptions) => {
         context.dispose()
       })
     }
+    process.on('SIGINT', () => process.exit())
   }).catch(() => process.exit(1))
 }
--- a/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
+++ b/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb
@@ -49,7 +49,9 @@ const importGlobPlugin = () => ({
         return; // Ignore unresolvable paths
       }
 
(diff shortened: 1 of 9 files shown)


sandstrom:
Awesome! 🎉

PR #763, opened 2023-05-21
DSD. You know it. You want it. You love it.

jaredcwhite:
**Update:** this branch now also includes #765 

Render [DSD](https://webkit.org/blog/13851/declarative-shadow-dom/) templates like a boss with the `dsd` helper / Liquid tag. 😎

Also provides `dsd_style` helper to inline CSS inside of a DSD template.

### Todo:

- [x] Pull in [Turbo Shadow](https://github.com/whitefusionhq/turbo-shadow) for the Turbo configuration so it polyfills DSDs
- [x] Docs
- [x] Tests
- [ ] Stretch goal: build (optional) Bridgetown layouts which use DSD right out of the gate. Defaults are important!


bridgetown-core/lib/bridgetown-core/utils.rb
```diff
@@ -523,6 +523,10 @@ def chomp_locale_suffix!(path, locale)
       end
     end
 
+    def dsd_tag(input)
```
ayushn21, line 526: How do you feel about allowing the `shadowrootmode` to be passed in?

```ruby
def dsd_tag(shadow_root_mode: :open, input)
  raise ArgumentError unless [:open, :closed].contains? shadow_root_mode
  %(<template shadowrootmode="#{shadow_root_mode}">#{input}</template>).html_safe
end
```

PR #779, opened 2023-07-06
Lock Rubocop to latest version that passes

michaelherold:
This is a 🔧 tooling change.

## Summary

This change stops ignoring the `Gemfile.lock` and checks it into the repository. This is a best practice to make contributing easier for new (and current) contributors.

One large paper cut is to get all of your changes to pass locally then have CI fail because of updates in the linting pipeline that aren't compatible with — in particular — code that you did not touch.

By checking `Gemfile.lock` in, it ensures that all contributors are working from the same versions of dependencies that the CI server is running.

In addition, this change also locks Rubocop to the newest version that relies on `rubocop-ast` <= 1.28.0. That version and 1.29.0 have a parsing bug in them that won't parse one of the files in Bridgetown's source code and error out. Also, newer versions fail due to new cops on source code that currently exists in the repository.

## Context

All recent PRs are failing due to issues that don't have to do with the changes within them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ output/
 .byebug_history
 .ruby-gemset
 *.gem
-Gemfile.lock
 
 # Files
 .analysis
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,253 @@
+PATH
+  remote: bridgetown-builder
+  specs:
+    bridgetown-builder (1.3.0)
+      bridgetown-core (= 1.3.0)
+
+PATH
+  remote: bridgetown-core
+  specs:
+    bridgetown-core (1.3.0)
+      activemodel (>= 6.0, < 8.0)
+      activesupport (>= 6.0, < 8.0)
+      addressable (~> 2.4)
+      amazing_print (~> 1.2)
+      colorator (~> 1.0)
+      erubi (~> 1.9)
+      faraday (~> 2.0)
+      faraday-follow_redirects (~> 0.3)
+      hash_with_dot_access (~> 1.2)
+      i18n (~> 1.0)
+      kramdown (~> 2.1)
+      kramdown-parser-gfm (~> 1.0)
+      liquid (~> 5.0)
+      listen (~> 3.0)
+      rake (>= 13.0)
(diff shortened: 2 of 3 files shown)


brandondrew:
Since it's normal to check it in, I feel compelled to ask:

- was there some problem that ignoring the Gemfile.lock was meant to solve?   
- does that problem remain unsolved?

michaelherold:
> was there some problem that ignoring the Gemfile.lock was meant to solve?

This is something that I used to do in my open source gems as well and was dissuaded from by the sage advice of @rafaelfranca. It stems from an [old article](https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) that argues that developing a library you want to be able to test on as wide a range of dependency combinations as possible to ensure that you're actually able to work with the range specified in your gemspec.

While this is at face value true, it ends up causing more headaches for your contributors than helping ensure compatibility. An alternative solution, if you _really_ want to do so would be to run CI after deleting `Gemfile.lock` and perhaps allow it to fail. That will (1) help show when a dependency fails its semantic versioning constraints or you don't have accurate semantic versioning constraints and (2) not block contributors (in particular new ones) from submitting PRs that are mergeable.

> does that problem remain unsolved?

Yes and no. Yes, it remains unsolved because of the edge case outlined above. No, because it's not something that really comes up. Rails has successfully checked in their `Gemfile.lock` [since 2015-02-18](https://github.com/rails/rails/commit/b1edc3789a39913b889b3ab18039ecca17835df3).

jaredcwhite:
@michaelherold @brandondrew Yeah, I don't see a particular reason not to commit it to the repo, I think it's just an historical anomaly at this point. And we _have_ had many headaches due to not having it.

If you don't mind updating it based on the latest state of `main`, then I can go ahead and merge. Thanks Michael!

michaelherold:
Done!

michaelherold:
Third time is the charm ... Done for real now! 😅

rafaelfranca:
Just to leave here the updated recommendation about Gemfile.lock in repositories:

https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html

PR #790, opened 2023-08-04
Fixed localization of paginated resources error

### LINKED ISSUE #789: Pagination and multi locales
When using pagination with `locales: multi`, I've run into the following error.

```
NoMethodError: undefined method `parent' for "blog/index.slim":String
/Users/jarrett/Projects/random/bridgetown/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb:16:in `block in all_locales'
/Users/jarrett/Projects/random/bridgetown/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb:15:in `select'
```

When digging further into this I realized that the `all_locales` method in Localizable tries to call `parent` on `relative_path`. When the `result_set` is a `Bridgetown::Resource::Base`, that works fine since it returns a `Pathname` object. When it is a `Bridgetown::GeneratedPage`, `relative_path` returns a `String` which breaks.

Now I'm not sure if the issue lies within `relative_path` because I'd assume a lot of other stuff would be broken if that wasn't supposed to return a String in this situation. Here is what I did to try and fix this.

```ruby
# frozen_string_literal: true

module Bridgetown
  module Localizable
    def all_locales
      result_set = case self
                   when Bridgetown::Resource::Base
                     collection.resources
                   when Bridgetown::GeneratedPage
                     site.generated_pages
                   else
                     []
                   end

      matching_resources = result_set.select do |item|
        if item.relative_path.is_a?(String)
          item.localeless_path == localeless_path
        else
          item.relative_path.parent == relative_path.parent
        end && item.data.slug == data.slug
      end

      matching_resources.sort_by do |item|
        site.config.available_locales.index item.data.locale
      end
    end

    def localeless_path
      relative_path.gsub("#{data.locale}/", '')
    end

  end
end
```

Now this worked fine for fixing the error and generating the proper pages. The problem that I still face is that the pages for anything, but my base locale are not returning anything when calling `in_locale(paginator.resources)`.

I'm not sure if I'm going down the wrong path trying to fix this so I figured it's time to open an issue. I also wasn't sure if this is related to #711.

**Bridgetown Version**: 1.3.0 "Kelly Butte"

**To Reproduce**
- Dynamically generate a resource using a builder and add_resource
- Make a page like `blog.erb` that defines `locale: multi` and `locales: ["en", "es"]`
- Enable pagination

jclusso:
This is a 🐛 bug fix. 

## Summary

I'm not sure if this is the right way to go about this. Additionally, there are no tests because I couldn't find any examples of testing paginate at all.

Fixes #789

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
+++ b/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
@@ -13,11 +13,7 @@ module Bridgetown
                    end
 
       matching_resources = result_set.select do |item|
-        if item.relative_path.is_a?(String)
-          item.localeless_path == localeless_path
-        else
-          item.relative_path.parent == relative_path.parent
-        end && item.data.slug == data.slug
+        matches_resource?(item)
       end
 
       matching_resources.sort_by do |item|
@@ -25,9 +21,16 @@ module Bridgetown
       end
     end
 
-    def localeless_path
-      relative_path.gsub("#{data.locale}/", '')
+    def matches_resource?(item)
+      if item.relative_path.is_a?(String)
+        item.localeless_path == localeless_path
+      else
+        item.relative_path.parent == relative_path.parent
+      end && item.data.slug == data.slug
     end
 
+    def localeless_path
+      relative_path.gsub("#{data.locale}/", "")
+    end
   end
 end
--- a/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
+++ b/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
@@ -30,7 +30,7 @@ module Bridgetown
     end
 
     def localeless_path
-      relative_path.gsub("#{data.locale}/", "")
+      relative_path.gsub(/\A#{data.locale}\//, "")
     end
   end
 end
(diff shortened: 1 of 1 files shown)


bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
```diff
@@ -13,12 +13,24 @@ def all_locales
...
+    def matches_resource?(item)
+      if item.relative_path.is_a?(String)
+        item.localeless_path == localeless_path
+      else
+        item.relative_path.parent == relative_path.parent
+      end && item.data.slug == data.slug
+    end
+
+    def localeless_path
+      relative_path.gsub("#{data.locale}/", "")
```
jaredcwhite, line 33: This feels too broad — it would mangle any URLs with, e.g., `/en/sheep/oxen/pigs` suddenly becomes `/sheep/oxpigs`

jaredcwhite (changes requested):

jaredcwhite:
I think you're on the right track…just need to tighten up the string sub.

bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
jclusso: Good call. How about I do this.

```suggestion
      relative_path.gsub(/\A#{data.locale}\//, "")
```

jclusso: I'll go with this fornow.

jaredcwhite (approved):

PR #791, opened 2023-08-08
Fix `link_to` when passed block

jclusso:
This is a 🐛 bug fix.

## Summary

- When passing a block to `link_to` the additional options were being dropped. 
- When passing a block to `link_to` in ERB, the block needs to be captured or the block is not inside the link element.


jclusso:
Not really sure what is causing the CI to still fail. 🤔

jaredcwhite:
Related: #794

jaredcwhite:
I'll go ahead and merge this, see if I can clear up any remaining CI issues after that. Thanks @jclusso!

PR #793, opened 2023-08-09
Add ActiveSupport::Inflector to Zeitwerk

jclusso:

## Summary

To make things easier I've made it so Zeitwerk uses ActiveSupport::Inflector out of the box. I'm not really sure how to add tests for this. Looking for assistance on that if required to merge this. I've testsed this in my own project and it works.

## Changes
- Configure Zeitwerk's inflector to use ActiveSupport::Inflector
- Add commented out example to config/initializer.rb
- Update existing documentation about ActiveSupport::Inflector to reference initializer configuration
- Add documentation for initializer

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/utils/loaders_manager.rb
+++ b/bridgetown-core/lib/bridgetown-core/utils/loaders_manager.rb
@@ -55,6 +55,7 @@ module Bridgetown
           next unless Dir.exist? load_path
 
           loader = Zeitwerk::Loader.new
+          loader.inflector = ActiveSupport::Inflector
           begin
             loader.push_dir(load_path)
           rescue Zeitwerk::Error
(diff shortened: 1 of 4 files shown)


jaredcwhite:
Thanks @jclusso, looks good. I need to think about the best way to merge this is as it _is_ a breaking change (I don't want folks to upgrade from v1.3 to 1.4 and suddenly various classes break). We may want to keep it opt-in only at the start, and then switch to default in v2.0. What do you think?

jclusso:
@jaredcwhite yeah, I didn't consider that it could be breaking. I think making it opt-in until v2 seems reasonable. Any suggestions on how you'd think we best do that?

jclusso:
I had some more thought on this, but not sure how you'd like to proceed. Instead of changing this PR, we can leave this here until v2.0 is ready and you can merge it then since I believe this is ready for that from a code and documentation perspective.

To make it optional, I could make another PR where we add `autoload_inflector` to `Bridgetown::Configuration`and I can update the documentation and initializer to set it.

jaredcwhite:
@jclusso That last suggestion sounds good to me. If it could be a configuration option to enable it, I'm all for it.

jclusso:
Replaced by #796 and #797

PR #817, opened 2023-10-10
Support `pipe` in Serbea 2.0, add new pure Ruby template syntax (WIP)

jaredcwhite:
Serbea 2.0 brings with it a new [pipe helper](https://www.serbea.dev/#add-pipelines-to-any-ruby-templates) that you can use in any Ruby context (including ERB). Example in a PORO:

```ruby
class PipelineExample
  include Serbea::Pipeline::Helper

  def output
    pipe("Hello world") { upcase | split(" ") | test_join(", ") }
  end

  def test_join(input, delimeter)
    input.join(delimeter)
  end
end
```

This PR brings that into Bridgetown templates and components, as well as adds a **brand-new** enhanced "HTML-in-Ruby" syntax based on heredocs. (_Update:_ I've extracted this out to a standalone gem, [Streamlined](https://github.com/bridgetownrb/streamlined), that Bridgetown will then import.)

Syntax example:

```ruby
class RubyHtmlText < Bridgetown::Component
  def template
    html -> { <<~HTML
      <p>This is #{text -> { "<b>escaped!</b> "}}</p>
      #{html "_yipee", -> { markdownify | sub("yipee", "yay") }}
    HTML
    }
  end
end
```

You can think of the new `text`, `html`, `render`, and `html_map` helpers for HTML-in-Ruby as somewhat analogous to "tagged template literals" in JavaScript. HTML-in-Ruby is an alternative template format to ERB and Serbea, and can be used in any `.rb` resource template as well as PORO components as in the example above.

A RuboCop linter will be provided so that any `HTML` or `MARKDOWN` heredoc with interpolated expressions not starting with an approved helper can be flagged. This is to ensure security against XSS…because Ruby doesn't provide any way to process interpolated expressions before they're inserted into the string, we have no way of HTML escaping content by default. By intentionally choosing the `text` helper, you get an escaped string. Or by using `html`/`render`/etc., it will not escape.

The file-based routes syntax is also improved, with a new Ruby front matter style available so you can write route handlers like:

```ruby
render_with do
  layout :page
  title "Access Your Account"
end
```

- [x] Tests
- [x] Extract linter to rubocop-bridgetown gem
- [x] Extract Ruby HTML template syntax to standalone gem, make dependency of Bridgetown
- [ ] Documentation
- [ ] **Breaking change** re: Roda cookies! (see comment)


jaredcwhite:
Note: this PR includes one **breaking change** to the default cookies setup for Roda (to set the cookie path to `/`…otherwise cookies sort of seem broken because they don't work across the whole site).

jaredcwhite:
I'm going to go ahead and merge this in, and create some follow-up issues for the documentation, etc.

Issue #609, opened 2022-08-04
feat: allow package manager selection upon install 


sandstrom:
## Summary

Allow selection of package manager (npm, yarn) during setup.

## Motivation

NPM, with all its flaws, is still the most popular package manager. Also, the package managers come and go. Having a standardized command for selecting one would make it easier to add more in the future.

## Guide-level explanation

- Remove or deprecate `--skip-yarn` and `--no-skip-yarn` flags.
- Add a `--js-package-manager` flag, which would accept `js-package-manager={npm,yarn,none}`. Selecting 'none' would be the same as the old `--skip-yarn` flag.
- The default behavior, when the flag is omitted, could still be to select yarn. It's not the default I would have chosen, but that's a minor thing, as long as changing is easy.

## Reference-level explanation

*Could add more details if this suggestion receives positive feedback.*

## Drawbacks

It's possible to remove yarn and install npm manually.

## Unresolved Questions

n/a

jaredcwhite:
Thanks for submitting this issue!

We're only ever going to support one JS package manager. There are many hooks in the system which expect a certain CLI to be available, so if it's not Yarn it would be something else exclusively. And a long-term goal of the project is to find a way to abstract this entirely and not rely on a system-installed package manger at all (aka switch to Ruby tooling for all high-level interfaces).

I appreciate the idea, and we'll keep an eye out for how this space continues to develop.

sandstrom:
Alright, if you're only going to pick one, I'd totally go with NPM. I know NPM has many flaws, but it's still the de-facto Node package manager. Also, with Github/Microsoft now running the NPM project, it's likely to remain well supported and stable for the foreseeable future.

Any default choice should be the one with the least friction. Everyone knows NPM and how it works (even though they may not have it as their preferred node manager). But not everyone knows how Yarn or pnpm works, so that's always going to be more of a threshold for newcomers.

***

As an aside, I understand the desire to abstract it away, but my guess would be that it's a lot of hassle. Better to not abstract too much over the front-end stuff, instead threat them as two separate processes, that produces different output (build artifacts).

***

Great work on BridgeTown, we're doing some tests now, to see if we can migrate our Middleman site over to Bridgetown.

sandstrom:
Another argument/anecdote for not abstracting over Yarn/NPM too much.

[...]

PR #821, opened 2023-10-29
auto-detect package manager in the plugin manager

KonnorRogers:

## Summary

Auto-detect how to install dependencies of plugins to improve the story of using any package manager

## Context

Related to #609 

This doesn't fix it. But this allows you to opt-out of yarn and use whatever package manager you want. We could also just bail if we can't find a `yarn.lock` or maybe check an ENV var like `NO_YARN_INSTALL` or something?

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/plugin_manager.rb
+++ b/bridgetown-core/lib/bridgetown-core/plugin_manager.rb
@@ -129,6 +129,22 @@ module Bridgetown
                               "Required #{required_gems.map(&:name).join(", ")}")
     end
 
+    def self.package_manager
+      if File.exist?("yarn.lock")
+        "yarn"
+      elsif File.exist?("package-lock.json")
+        "npm"
+      elsif File.exist?("pnpm-lock.yaml")
+        "pnpm"
+      end
+    end
+
+    def self.package_manager_install_command
+      package_manager == "npm" ? "install" : "add"
+    end
+
+    # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
+
     # Iterates through loaded gems and finds yard-add gemspec metadata.
     # If that exact package hasn't been installed, execute yarn add
     #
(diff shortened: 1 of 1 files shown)


sandstrom:
Just to chime in, my suggestion would be that Bridgetown should be less concerned about the JS ecosystem and package managers.

Instead, it should provide generic primitives (hooks, etc) that any package manager can use.

I've explained my thinking in these two posts:
- https://github.com/bridgetownrb/bridgetown/issues/609#issuecomment-1209642057
- https://github.com/bridgetownrb/bridgetown/issues/609#issuecomment-1221277335

jaredcwhite:
@sandstrom Yeah I think we could easily extract all this stuff out to a more third-party-configurable module of some sort rather than buried in a core class.

jaredcwhite:
I want to get this merged in the meantime though. I think @KonnorRogers we just need the linting fixed and then we're good to go!

sandstrom:
> I want to get this merged in the meantime though. 

I understand, sounds good!

> Yeah I think we could easily extract all this stuff out to a more third-party-configurable module of some sort rather than buried in a core class.

The whole ecosystem for JS/node moves fast, and trying to keep up is impossible.

Apart from the comments cited above, Middleman is another good example.

https://middlemanapp.com/advanced/external-pipeline/#webpack-example

They have the concept of an 'external pipeline' and try not to worry too much about JS/CSS/etc.

Bridgetown can still provide some helpful documentation on how to set it up, and even have the 'start new project' CLI tool generate some boilerplate code that set things up with webpack or esbuild.

But I think the framework should avoid trying to write a wrapper API around all the potential functionality that may go into an asset pipeline, and instead leave it to node-land tools.

KonnorRogers:
@jaredcwhite fixed

jaredcwhite:
@KonnorRogers Fantastic! I hope to cut another point release soon, so this will be a welcome improvement.

@sandstrom I think that's generally our position, but it's also easier to maintain tooling over time that's part of the framework's "runtime" than something that's part of a starter kit or whatever. That's why we don't simply generate a single esbuild config file and then wipe our hands forevermore, but instead the main portion of the config is part of a file "owned" by the framework that can get auto-updated. As someone who starts and operates dozens of Bridgetown sites, I take comfort knowing the esbuild config can always be easily brought up-to-date across all of them.

PR #824, opened 2023-11-02
Update Netlify configuration

### LINKED ISSUE #823: Netlify bundled configuration issues
I'm using Netlify and noticed some issues that I wanted to bring up because I'm not sure about them. The CSS and JS path appear to be wrong. My CSS and JS do not appear to be nested in a `css` and `js` folder, but rather directly in the `static` folder. (ex: `output/_bridgetown/static/`)

https://github.com/bridgetownrb/bridgetown/blob/9b769d55fa47db515a62af338b5fc3281212d1c0/bridgetown-core/lib/bridgetown-core/configurations/netlify/netlify.sh#L6-L7

Additionally, my build does not appear to have a `NETLIFY_IMAGES_CDN_DOMAIN` and I was wondering where this came from. I can't find much about this.

https://github.com/bridgetownrb/bridgetown/blob/9b769d55fa47db515a62af338b5fc3281212d1c0/bridgetown-core/lib/bridgetown-core/configurations/netlify/netlify.sh#L14

@ayushn21 maybe you can provide some insight since I saw you authored this.

jclusso:
This is a 🐛 bug fix.

## Summary

- CSS and JS path are incorrect
- `NETLIFY_IMAGES_CDN_DOMAIN` doesn't appear to exist anymore and isn't documented (https://docs.netlify.com/configure-builds/environment-variables/#read-only-variables)
- Netlify removed all post processing besides `pretty_urls`

Fixes #823

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/configurations/netlify/netlify.sh
+++ b/bridgetown-core/lib/bridgetown-core/configurations/netlify/netlify.sh
@@ -3,12 +3,11 @@
 
 echo "Updating netlify.toml with references to our built files"
 
-CSS_PATH=`find output/_bridgetown/static/css/*.css -type f | sed -e 's,output\/,/,g'`
-JS_PATH=`find output/_bridgetown/static/js/*.js -type f | sed -e 's,output\/,/,g'`
+CSS_PATH=`find output/_bridgetown/static/*.css -type f | sed -e 's,output\/,/,g'`
+JS_PATH=`find output/_bridgetown/static/*.js -type f | sed -e 's,output\/,/,g'`
 
 echo "CSS Path: ${CSS_PATH}"
 echo "JS Path: ${JS_PATH}"
 
 sed -i s,CSS_PATH,${CSS_PATH},g netlify.toml
 sed -i s,JS_PATH,${JS_PATH},g netlify.toml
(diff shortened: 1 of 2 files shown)


bridgetown-core/lib/bridgetown-core/configurations/netlify/netlify.sh
```diff
@@ -3,12 +3,11 @@
 
 echo "Updating netlify.toml with references to our built files"
 
-CSS_PATH=`find output/_bridgetown/static/css/*.css -type f | sed -e 's,output\/,/,g'`
-JS_PATH=`find output/_bridgetown/static/js/*.js -type f | sed -e 's,output\/,/,g'`
+CSS_PATH=`find output/_bridgetown/static/*.css -type f | sed -e 's,output\/,/,g'`
```
ayushn21, line 6: @jaredcwhite is this path universal irrespective of front-end bundling option? Or is it specific to esbuild only?

ayushn21 (approved):
Just need to double check with Jared on that one comment but looks good to me otherwise. 

If that path isn't universal across all our front-end bundling options, I recommend we remove the preload headers from this template ... they're really not that critical and more of a nice-to-have.

PR #839, opened 2024-02-17
Improve Netlify bundled configuration

jclusso:
## Summary

- Add support for using `netlify dev`. This is important if you want to test redirects or use functions.
- Removed Link headers. This was breaking stuff locally and Netlify doesn't have a way to apply headers conditionally without using scripts. Honestly, I'm not sure it's really that important anyway. I checked a bunch of sites including Google, Cloudflare, Netlify, Apple, Stripe, and more. None of them set these headers. This simplifies stuff to remove them.


jaredcwhite:
I'm guessing you'd need to remove the references to the shell script in `bridgetown-core/lib/bridgetown-core/configurations/netlify.rb` as well?

jclusso:
> I'm guessing you'd need to remove the references to the shell script in `bridgetown-core/lib/bridgetown-core/configurations/netlify.rb` as well?

Whoops, missed that.

PR #842, opened 2024-02-18
Improve localization helpers

### LINKED ISSUE #837: feat: `_html` translation key automatic safety support
## Summary

Translation keys that end with `_html` would be nice if they automatically were marked safe so we can avoid using `==` or `.html_safe` in views. This is something that happens in Rails and I think it would be nice to have this feature. Going to make a PR for this when I have some time, but wanted to document it here and see if there are any objections.

jclusso:
- Added support for relative translation keys (ex: `t('.foo')`).
- Added support for automatically marking translation keys ending with `_html` as `html_safe`.
- Renamed methods to `translate` and `localize` with aliases of `t` and `l`.
- Added tests.

Closes #837

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/helpers.rb
+++ b/bridgetown-core/lib/bridgetown-core/helpers.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'active_support/html_safe_translation'
+require "active_support/html_safe_translation"
 
 module Bridgetown
   class RubyTemplateView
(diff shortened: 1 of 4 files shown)


jaredcwhite:
This looks really good @jclusso!

I would like to avoid adding the dependency on Active Support here (since we're generally headed in the opposite direction). Looking at the code being called, it seems fairly self-contained:

https://github.com/rails/rails/blob/main/activesupport/lib/active_support/html_safe_translation.rb

If you're feeling particularly ambitious, I might ask you to write a greenfield take on this functionality—otherwise, I'm completely fine with vendoring this into its own module file somewhere and a commented link to the original code on GH. LMK what you think, I'm good with either option. TY 🙏🏻

jclusso:
Thanks @jaredcwhite!

I'm not super familiar with how the whole string safety thing works and I wouldn't feel comfortable blindly ripping this out of Active Support. I feel it's also worth noting that Active Support is [one of the top 10 most downloaded gems](https://rubygems.org/stats) and it really seems like a poor use of my time to reinvent the wheel. I personally can't justify that since I have a limited amount of time I can put into OS projects and I much rather put it into something that moves the needle. I hope you can understand.

It would be a shame if that was a deal breaker for this PR since I think a lot of people would find this useful.

jaredcwhite:
I'll go ahead and merge and rewrite that dependency. We're just not going to be relying on any new Active Support functionality going forward.

Thanks for the PR.

PR #848, opened 2024-02-24
Fix relative translation keys for views with dash

jclusso:
When your view has a dash in the name, the key is getting a dash in it as well. I think it's likely that most times a dash in the view is for nice URLs and having to mix dashes with underscores in your translation keys is not ideal. This PR replaces all dashes in the view name with underscores.

This is a 🐛 bug fix.

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/helpers.rb
+++ b/bridgetown-core/lib/bridgetown-core/helpers.rb
@@ -158,7 +158,7 @@ module Bridgetown
 
         if key&.start_with?(".")
           view_path = view&.page&.relative_path&.to_s&.split(".")&.first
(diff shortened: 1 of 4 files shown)


jclusso:
After writing this PR, I'm second guessing if this is a good idea. I'm running into relative key identification using `i18n-tasks` because of this. It's looking for the dashes because of the file name. I'm torn because I dislike having the dashes in the translation keys.

jclusso:
So I've modified my site to use `permalink` in the frontmatter to set the friendly dash URL while having underscores in the filename. @jaredcwhite let me know your thoughts before I axe this this.

jaredcwhite:
Yeah, I think we should leave as-is…unless there's a specific technical reason to push us in this direction, I don't know that we should be messing with dashes vs. underscores in the translation keys.

Issue #785, opened 2023-07-15
Add note on blog post about tailwindcss


jclusso:
When searching for tailwindcss the first thing that you'll find in Google and the Bridgetown site search is [this outdated blog post](https://www.bridgetownrb.com/news/postcss-tailwind-plugins-oh-my/). This led me down a rabbit whole and it wasn't until later that I found https://github.com/bridgetownrb/tailwindcss-automation linked from the [Bundled Configurations](https://www.bridgetownrb.com/docs/bundled-configurations) page. A little note at the top of that blog post would've saved me some confusion and probably will help others. I was going to submit a PR, but blog posts are a bit personal so I figured it was best to just open an issue.

PR #849, opened 2024-02-24
Add notice to Tailwind CSS blog post about bundled configuration

jclusso:
Closes: #785


jaredcwhite:
@jclusso Thanks for this! Just a quick note re: terminology, "bundled configurations" are automations which are bundled in the core gem. Since this community-maintained automation isn't bundled, it should be referred to as an "automation" rather than "bundled configuration". Other than that, sounds good!

jclusso:
@jaredcwhite I see. I've made that change. I know understand why my wording was incorrect since it previously was a "bundled configuration". I had stole some of the wording from below.

https://github.com/bridgetownrb/bridgetown/blob/f2f259423f1ac73bcb65452c5344090a6506b833/bridgetown-website/src/_docs/bundled-configurations.md?plain=1#L42-L46

PR #863, opened 2024-03-07
esbuild: Add webp to the file endings loaded as files

moonglum:

## Summary

This PR adds a few file extensions as a file type to the list in `esbuild.defaults.js`:

* `.jpeg`
* `.avif`
* `.jxl`
* `.webp`

With this change, [WebP](https://en.wikipedia.org/wiki/WebP), [AVIF](https://en.wikipedia.org/wiki/AVIF) and [JPEG-XL](https://en.wikipedia.org/wiki/JPEG_XL) images are supported in the same way that PNG and JPEG are supported. It also adds `.jpeg` as the alternative file extension for JPEG files.

## Context

[This is from a discussion about me failing to fingerprint images](https://github.com/bridgetownrb/bridgetown/discussions/835)

### DIFF
--- a/bridgetown-website/config/esbuild.defaults.js
+++ b/bridgetown-website/config/esbuild.defaults.js
@@ -318,9 +318,13 @@ module.exports = async (esbuildOptions, ...args) => {
     bundle: true,
     loader: {
       ".jpg": "file",
+      ".jpeg": "file",
       ".png": "file",
       ".gif": "file",
       ".svg": "file",
+      ".avif": "file",
+      ".jxl": "file",
+      ".webp": "file",
       ".woff": "file",
       ".woff2": "file",
       ".ttf": "file",


matiaskorhonen:
Sort of related: `.jpeg` should be added as well

I ran in to that recently but never got around to opening a PR…

moonglum:
Thanks, @matiaskorhonen :+1: Added it, alongside AVIF and JPEG-XL. That would then complete the list of the image formats I have on my radar.

jaredcwhite:
Fantastic!

matiaskorhonen:
Should this change have actually gone into `bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb` instead?

moonglum:
I think you're right, @matiaskorhonen . PR is here: https://github.com/bridgetownrb/bridgetown/pull/870

PR #864, opened 2024-03-10
Relax the Rouge version requirement

matiaskorhonen:
This is a 🐛 bug fix.

## Summary

Relax the Rouge version requirement from `~> 3.0` to `>= 3.0, < 5.0` to allow Rouge 4.x to be used. Rouge 3.x can still be used if required.

**Rouge changelog:** https://github.com/rouge-ruby/rouge/blob/master/CHANGELOG.md

There shouldn't be any breaking changes aside from a single lexers that has been removed and support for Ruby 2.6 and below has been dropped. On the upside there are new lexers and CSS themes.

## Context

No related issues.


jaredcwhite:
@matiaskorhonen Sounds good!

PR #865, opened 2024-03-16
Use Global Regex for Stimulus Configuration

MSILycanthropy:
This is a 🐛 bug fix.

## Summary
When using the Stimulus Configuration, the generated code for loading Stimulus controllers makes a couple `replace` calls to translate a path to a Stimulus identifier. 

This means that if you have a controller whose filename is `do_a_thing_controller.js`, this gets translated into the Stimulus identifier `do-a_thing`, when one would expect it to have been `do-a-thing`. There case for controllers nested in directories is similar.

With the way Stimulus currently works, it's incredibly annoying to debug this, because there are no warnings. Your controller just seemingly _doesn't work_.

## Context
When building my personal site with Bridgetown and Hotwire, I added a controller that used multiple underscores, and then spent an embarrassing amount of time trying to figure out what I had done wrong. Was surprised to see that only one replacement had taken place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Use global regex for Stimulus Configuration
 
 ## [1.3.2] - 2024-01-01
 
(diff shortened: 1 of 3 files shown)


jaredcwhite:
Thanks @MSILycanthropy! FYI I took the liberty of committing a fix for the regex backslash so it carries over to the JS code from the Ruby heredoc. I know, confusing! (Otherwise the JS file would just have `///` istead of `/\//`.)

PR #870, opened 2024-03-18
Add image formats to esbuild.defaults.js.erb

moonglum:

## Summary

Follow up to https://github.com/bridgetownrb/bridgetown/pull/863#issuecomment-2003220088
This adds support for additional image file extensions to the bridgetown esbuild configuration.

## Context

@matiaskorhonen pointed out that the previous change missed the change to the `esbuild.defaults.js.erb` file.


jaredcwhite:
Got it…I'll cut a patch release shortly!

Issue #860, opened 2024-03-06
Remove hard dependency on Puma and add generic Rack compliance


ayushn21:
## Summary

We took a step forward by dropping WEBrick for Puma, and I think we should take this further and remove this hard dependency as well. Roda is a Rack-compliant framework and there's no technical reason why we need to depend exclusively on Puma.

We can add compliance to Rack in general and have a sensible Puma setup out of the box, but if someone wanted to bring their own server, they should be able to.

## Motivation

This is a result of an ADHD fuelled deep dive into Rack 🤣. These two talks led me down this rabbit hole:

https://www.youtube.com/watch?v=x8qOIY252sw
https://www.youtube.com/watch?v=yXyj9wlkJKM

Falcon is a new Rack compliant web server built on top of the Async gem. This makes it easy to write heavily concurrent apps, as long as the performance bottleneck is with IO. Puma uses threads and waiting for server side API requests to come back would hold up the thread meaning it couldn't serve other requests. This could be a problem on small deployments. 

Falcon uses Ruby Fibers for each request which are very cheap and the Async gem is very efficient at switching between them when they're blocked by IO. I think this might be a great option for Bridgetown somewhere down the line. 

In any case, I think folks should be able to choose any Rack server since live deployments of Bridgetown are also now possible, and the server isn't just for development.

## Reference-level explanation

We can use the `rackup` gem to start a generic Rack server in the `start` command. Since Puma is already Rack compliant, I don't foresee any huge problems with this.

We should also bump our minimum Rack requirement to v3. Without a massive legacy, I don't see why we shouldn't. Rack 3 added native support for bi-directional communication. This was quite hacky to do in v2. Rack 3 also makes significant architectural changes wherein `rackup` was broken out into its own gem. It would be cleaner to just go all-in on v3 instead of trying to support both. 

Since Bridgetown is primarily used in static deployments, I don't foresee any issues with this. Even for users of live deployments, the change should seamless.

Falcon is also a fully Rack 3 compliant and HTTP/2 compliant server making it a viable option for folks who would like to use it.

## Drawbacks

I don't think there's any reason not to.

PR #875, opened 2024-04-07
Refactor start command to use Rackup instead of Puma

ayushn21:

## Summary

This PR refactors the `start` command to use `Rackup` instead of Puma. This way we can run Bridgetown with any Rack compliant server and aren't restricted to Puma.

Rackup will automatically detect from the `Gemfile` and start it. Our out-of-the-box Puma setup still works as is.

To do:

- [x] Remove/amend usages of `options[:using_puma]`.
- [x] Remove hardcoding of host/port and use configuration.
- [x] Move "Now serving at:" logging to `start` file and use info from Rackup.
- [x] Test with another server like Falcon.
- [x] Test without a server in the Gemfile (Should start with WEBrick I think).
  -  Starts with WEBrick, but live reload causes errors. I think we can ignore this since we setup Puma by default.
- [x] Generic testing.

## More details

I had to refactor the way we start the Bridgetown build and aux processes to make it work with this setup. We're now forking a process which runs the build in the background. That in turn starts the aux processes like the frontend watcher. As such I now track PIDs in files so we can kill the processes we start despite stuff happening in different processes.

## Context

Fixed https://github.com/bridgetownrb/bridgetown/issues/860


jaredcwhite:
@ayushn21 I spent a little time with this and it feels pretty solid! Excited to see the progress.

ayushn21:
I've disabled live reload when using servers other than Puma. It's causing a few issues. Probably a can we can kick further down the road I think.

ayushn21:
Hey @jaredcwhite could you stress test this PR a bit please? I'm too close to it to test it properly 🙈

jaredcwhite:
I think it's looking really great @ayushn21. 🙌 I have a couple of small fixes on hand regarding when the frontend process isn't running (it crashes on Ctrl+C due to missing PID) and just some process messaging on startup. Mind if I push up and we can do a final review, or do you want to step through it ahead of time?

jaredcwhite:
I took the liberty of pushing up one fix, as there was a bug where the watcher in SSR mode wasn't returning from a method before setting up an INT trap, causing automated tests to hand and run indefinitely!

ayushn21:
> Mind if I push up and we can do a final review, or do you want to step through it ahead of time?

Go ahead and push it :).

> I took the liberty of pushing up one fix, as there was a bug where the watcher in SSR mode

Awesome thanks! I was stumped by that when I was working on this but was too tired to figure it out at the time so I kicked the can down the road. Thanks for fixing it!

jaredcwhite:
@ayushn21 All right, I pushed up a couple of things…binding to `0.0.0.0` instead of `localhost` means other devices on the same network can access the site (I use this all the time, like for testing on my phone), and only stopping the frontend bundling if it's actually running (and not in production). Also still want to set the process name so when I run `ps` I can see a nice readout for the Bridgetown process, next to Puma.

PR #882, opened 2024-04-15
Significant refactor of the Routes plugin

### LINKED ISSUE #736: SSR Home page (index.html)
Server-side rendering of the homepage does not seem possible? I've followed the docs and tried the file-based routing. The [items example](https://edge.bridgetownrb.com/docs/routes#file-based-dynamic-routes) works as expected.

**Bridgetown Version**:

bridgetown 1.2.0 "Bonny Slope"

**To Reproduce**

**Create a new site and enable bridgetown-routes**

1. Create new bridgetown site
2. In `config/initializers.rb` un-comment `init :"bridgetown-routes"`
3. In the `Gemfile` un-comment `gem "bridgetown-routes"`
4. Run `bundle install`

**Create the route**

1. Create the routes directory `src/_routes`
2. Create `index.erb` inside of routes
3. Use code from items example

**Current behavior**

When you visit `http://localhost:4000` you get a 404

image

**Expected behavior**

The page to be rendered

jaredcwhite:
There was some not-particularly-attractive code lurking in the file-based routes plugin. It had been brute-forced to work, and then didn't receive the necessary spit 'n' polish it deserved. This PR provides that.

- More modular and OOP-like code
- Better Roda plugin layout
  - We're not using Roda's `before` hook anymore, which means end-user developers can leverage it if they so choose
- Downstream plugins can leverage file-based routing themselves
- Dynamic `index` route is now supported instead of static `index.html`

That last one is huge, as it means a Bridgetown + Roda site could effectively be a completely dynamic Ruby application! And it closes #736.

Also moved previously monkey-patched Roda Public code into a streamlined standalone SSG plugin (to be extracted out later — it's really just a small wrapper around `Rack::Files` as was the earlier Public plugin), and now it's serving static `index.html` using the same `Rack::Files` server.


jaredcwhite:
I should mention here: a longer-term goal is to extract most of this functionality out into a standalone Roda plugin which would be view layer-agnostic, so it could work even with Roda's own `render` plugin. Then any Roda app developer could leverage the file-based routing convention.

PR #885, opened 2024-04-23
Fix typo

soulcutter:

## Summary
Surely "tast" should be "task" :)

## Context
I cannot let a typo exist if there's something I can do about it! 😉


ayushn21:
Thanks :D

PR #887, opened 2024-04-27
Use bundler's binstub for bt shortcut

### LINKED ISSUE #852: docs: use symbolic link for `bt` binstub
## Motivation

I don't think this is technically a bug or feature request so I used docs. I would propose modifying `bt` binstub with symbolic link.
Now `bin/bridgetown` and `bin/bt` are the same commands. Since bundler generates `bin/bridgetown` so bundler can update the binstub technically (or other tools). Using a symbolic link would allow tools like bundler to make such changes.

## Suggestion

This means a change in `bridgetown-core/lib/bridgetown-core/commands/new.rb`.

```ruby
          Bridgetown.with_unbundled_env do
            inside(path) do
              run "bundle install", abort_on_failure: true
              run "bundle binstubs bridgetown-core"
              # bundle's binstub is a wrapper to call the actual executable in bridgetown-core's executable.
              # Use symbolic link to allow bundler or other tools to change binstubs reliably.
              run "cd bin && ln -s bridgetown bt"
            end
```

erickguan:

## Summary

I tested this locally and I can use `bin/bt` to call `bridgetown-core/bin/bridgetown`. Please comment if I need to make changes.

## Context

Fixes #852


jaredcwhite:
Looks good @erickguan!

PR #911, opened 2024-08-28
[website] Add Kaigi on Rails website to the list

okuramasafumi:
## Summary

This PR adds Kaigi on Rails website to the list in the community page.

## Context

I just noticed the list, and since we use Bridgetown this year (thank you!), I thought it's worth adding it.


ayushn21:
Thanks @okuramasafumi :D

PR #915, opened 2024-09-06
Memoize resource URLs for major performance gain

### LINKED ISSUE #914: Memoize relative_url for performance gains
My 250 pages site has a sidebar which links to every/most pages. I was tired to building the site taking so long, so after a bit of profiling, I noticed that `#relative_url` isn't memoized. So every link needs to rebuild the same relative urls.

Memoizing with the following in my `config/initializers.rb` made my build time go from 10s to 5s.

```
module MemoizeRelativeUrl
  def relative_url
    @relative_url ||= super
  end
end
Bridgetown::Resource::Base.prepend(MemoizeRelativeUrl)
```

I don't know if a relative_url is meant to change over the course of a build, but if it isn't, this is a massive usability improvement for me.

I don't know if the change should be on `Bridgetown::Resource::Base` or `Bridgetown::Resource::Destination` or somewhere else, but I think there should definitely be memoization somewhere.

jaredcwhite:
Resolves #914

_Awesome_ detective work by @MaxLap, we're able to get full build performance gains of 15-25% or maybe even higher! Apparently the code to determine relative/absolute URLs for resources is somewhat expensive, and repeating that constantly throughout template rendering really adds up. This PR memoizes the values upon resource read, and resets them if there's a fast refresh (in case the permalink was changed or whatever).


MaxLap:
Glad you are seeing benefits too!

A bit pedantic, but the way this is implemented feel more like a pre-calculation than a memoization ;). I don't know the overall flow of the system, but since the place where you set the instance variables is not the place where you produce the value nor the place where you need it, it feels like the values could sometimes be calculated for nothing. 

Visually, using ||= directly in the relative_url and absolute_url is a much more common pattern. So much so I thought it was a typo when I first say it.

jaredcwhite:
@MaxLap You're right, I was experimenting with different patterns and testing the outcome, and I landed on that one without it needing to be like that. I've updated to `||=`. 🙏

Issue #946, opened 2024-11-26
feat: Bits and bobs to fulfil complete Rack compliant


ayushn21:
## Summary

With Bridgetown v2.0, we added the `rackup` gem instead of instantiating Puma directly. This paved the way for the user to swap web server.

There are still a couple of issues to resolve before we can claim to be fully Rack compliant and support all Rack web servers. This issue tracks the remaining work.

## Motivation

There's no reason not to. We inherited a dependency on WEBrick from Jekyll, and replaced it with Puma. A step in the right direction, but we should go all the way. Roda is Rack-compliant. There's no technical reason for Bridgetown not to be.

## Checklist

- [ ] Ensure Bridgetown can be started with `bundle exec puma` and `bundle exec falcon serve`.
- [ ] A user should be able to setup Falcon easily. We should offer an out of the box setup in the future, but one step at a time.
- [ ] Remove the `require "puma"` dependency in the console command: https://github.com/bridgetownrb/bridgetown/blob/7f48f7ec2cd939967285face726a6d495a321137/bridgetown-core/lib/bridgetown-core/commands/console.rb#L83
- [ ] Ensure the live dev reload code works with all web servers.

PR #948, opened 2024-11-27
Require bridgetown when booting from Rack

ayushn21:
In order to start Bridgetown by running the web server directly (`bundle exec puma`), we need to `require "bridgetown"` in the boot file.

Without it, we get all sorts of uninitialized constant errors.  

This is one of those fixes where I'm not sure what I'm doing, and whether it'll have any bad consequences. I just hammered it until it worked and seems to be working fine in my tests. 

@jaredcwhite is this a valid fix? Please don't merge it until we know for sure it's not gonna make things catch fire haha.

Part of https://github.com/bridgetownrb/bridgetown/issues/946

### DIFF
--- a/bridgetown-core/lib/bridgetown-core.rb
+++ b/bridgetown-core/lib/bridgetown-core.rb
@@ -133,10 +133,16 @@ module Bridgetown
 
     # Set up the Bridgetown execution environment before attempting to load any
     # plugins or gems prior to a site build
-    def begin!
+    def begin!(with_config: :preflight)
       ENV["RACK_ENV"] ||= environment
 
-      Bridgetown::Current.preloaded_configuration = Bridgetown::Configuration::Preflight.new
+      if with_config == :preflight
+        Bridgetown::Current.preloaded_configuration ||= Bridgetown::Configuration::Preflight.new
+      elsif with_config == :initializers &&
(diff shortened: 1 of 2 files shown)


jaredcwhite:
@ayushn21 Off the top of my head, I believe if we're completely bypassing Bridgetown's own CLI, we'd need to replicate calling `Bridgetown.begin!` which loads up the right Bundler context. Not sure where that could go exactly, and we'd want to make sure we don't accidentally call it twice.

ayushn21:
@jaredcwhite ah ok cool, I'll have a poke around.

ayushn21:
What if we say that running the web server directly won't trigger a site build ... it'll only run Bridgetown in a live context ... i.e. Roda routes + anything already in the `output` folder ... do we still need to call `Bridgetown.begin!` then?

jaredcwhite:
@ayushn21 Yeah, the server/Roda context would still need the same bootup sequence. Running a build process is just what could happen _after_ initial boot. See `Bridgetown::PluginManager.setup_bundler` for an example.

I think we may need some kind of additional global state to indicate if the boot sequence has already run…if not, the rack boot needs to kick it off immediately after `require "bridgetown"`. I think…

PR #954, opened 2024-12-08
Revamp the `minitesting` bundled configuration so it uses `Rack::Test`

### LINKED ISSUE #570: feat: support Roda routes testing within the `minitest` configuration
aka setup the infrastructure to use Rack::Test, Capybara, Cuprite, etc.

(Note to self: need to revisit the experiments in the `rodanew` project and extract)

jaredcwhite:
At this juncture, I'd say it really makes the most sense to do automated testing at the Rack level so you can test both static routes (post site build) and dynamic routes, all via the Roda server application under the hood.

This PR updates the `minitesting` bundled configuration, as well as adds a canonical boot sequence and Minitest-based `Bridgetown::Test` superclass that test files can subclass. I'm backing out of "shoulda" etc. in favor of standard test definitions, and the old Rails DOM Assertions stuff is gone entirely in favor of Nokolexbor, mirroring how the HTML Inspectors feature works. I think this is much cleaner overall, and provides for really nice static+dynamic test ergonomics.

We'll eventually want to double-check this works nicely with the [Bridgetown Sequel](https://github.com/bridgetownrb/bridgetown_sequel) gem as well.

Resolves #570

**Addendum:** I also added support for Minitest expectations to Bridgetown tests — that's the syntax where you can say `expect([1,2,3]).must_include(2)`. I'm starting to like that more than `assert_` now that I've been spending time with it. In fact, I went a step further with some custom extensions so you can write "intuitive" expectations like `expect(something) == "to equal this"`. Whoa!

- [x] Add docs for the expectations to Testing page

As part of these changes, I removed the RSpec Mocks library and replaced with Minitest mocks and stubs.

**Update 2:** Well, this has become _quite_ the rabbit hole. I ended up extending the Spec DSL fully into the base test classes, allowing the use of `describe`, `before`/`after`, and `it`. This effectively lets us migrate away from the "shoulda" gem, because `context`, `setup` and `should` essentially map 1:1 with the Spec DSL. Once we transition, we can remove the shoulda gem dependency entirely and rely on "vanilla" Minitest functionality. 😎


jaredcwhite:
Ooo, this is worth looking into:

https://github.com/kematzy/minitest-rack

PR #973, opened 2025-02-09
Update Pagination Documentation

brainbuz:
Add most of the paginator methods, for both paginating a collection and using the paginator in templates. 

There were a few attributes I skipped because they did not appear to be things an end user would want to touch, these include:  documents (is this redundant with collections/resources?), page_trail/trail (is this a feature that isn't documented or not fully implemented?), indexpage (setting this to any other page broke the page), extension (ignored if indexpage is nil), debug (not sure this is useful outside of plugin development)

## Summary

Review 

## Context

No Github issues. 

I had to look at the source when figuring out how to use the paginator, and am responding to the call for documentation review in the doc.


jaredcwhite:
Thank you so much @brainbuz!

PR #974, opened 2025-02-10
doc update. sass deprecated @import, @use should work in place.

brainbuz:

## Summary

sass deprecated the @import function. the @use function should work in its place.  [@import in sass docs](https://sass-lang.com/documentation/at-rules/import/), [sass deprecation blog entry](https://sass-lang.com/blog/import-is-deprecated/)

## Context

No github issues, just deprecation warning to avoid.

### DIFF
--- a/bridgetown-website/src/_docs/frontend-assets.md
+++ b/bridgetown-website/src/_docs/frontend-assets.md
@@ -70,7 +70,7 @@ npm install name-of-css-framework
 And then add:
 
 ```css
-@import "~css-framework/css-framework";
+@use "~css-framework/css-framework";
 ```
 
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Thanks @brainbuz for pointing that out. I'm wondering if you can verify one of those examples in a live project and report if it works? I don't want to update the docs with the new syntax if there's some sort of error (I must confess I don't personally use Sass these days and I've never used `@use`.)

brainbuz:
The @import statements trigger deprecation warnings from Dart Sass. Their recommendation is @use instead. Doing this got rid of the deprecation warnings and SASS is working for me.

PR #980, opened 2025-03-04
docs: add support for tabbed multiple language code examples

jaredcwhite:
This PR adds a component which lets ERB and Liquid code examples live within a tabbed interface. We can improve on the component with flexibility as to which languages are demonstrated and even have more than two, but this is supporting the main use case.

DEMO: https://bridgetown-beta-pr-980.onrender.com/docs/resources


bridgetown-website/src/_components/documentation/multilang.rb
```diff
@@ -0,0 +1,20 @@
+class Documentation::Multilang < Bridgetown::Component
+  def template
+    split_content = content.split("===")
+
+    html -> { <<~HTML
+      <sl-tab-group>
+        <sl-tab slot="nav" panel="erb">ERB</sl-tab>
```
surrim, line 7: What about outsourcing this code as a [Bridgetown plugin](https://www.bridgetownrb.com/plugins) with user-defined languages? This would be great for many projects :)

jaredcwhite: I would love that! I think that can easily be done as a follow up a little down the road, or someone else's welcome to beat me to the punch. ☺️

Issue #883, opened 2024-04-21
docs: /docs/content/pagination


sjosephk1975:
## Motivation

Me as a beginner :-) was not 100% sure, which file should be updated.

## Suggestion

Mention that bridgetown.config.yml should be updated. That is all.

brainbuz:
A lot has been done on pagination documentation.

the current documentation shows for ruby templates paginator.each, but my experience is that paginator.resrouces.each is what works.

Once you update to make live behavior match that doc, this one looks goog to close.

Issue #807, opened 2023-09-06
Pages which `paginate` are excluded from `collections.pages`


goulvench:
I have pages in the `src/_pages/` folder which I list in a navigation partial. I fetch each page using `collections.pages.resources_by_slug` to access its title and URL. This is very practical because Bridgetown automatically gives me the localised title and URL (the website is generated in 2 locales).

However, when I add pagination in the frontmatter of a page, it disappears from `collections.pages`.

**Bridgetown Version**:
bridgetown 1.3.1 "Kelly Butte"

goulvench:
@jaredcwhite Good point! I was far from having 2 pages' worth of posts, so I hadn't noticed.

In the end I ditched pagination, the list page will list all posts, and individual posts have links to the previous/next post in_locale, which doesn't require pagination.

jaredcwhite:
Use case for why this is still relevant: https://github.com/bridgetownrb/bridgetown/pull/987#issuecomment-2755789948

jaredcwhite:
Not sure I can come up with a workaround off the top of my head, but I'll give this some thought.

PR #987, opened 2025-03-20
add initializers.rb examples to match legacy yaml examples

brainbuz:
I've gone through the doc looking for examples in legacy yaml and added equivalent initializers code with multilang. configuration.md had both examples, but they were not equivalent, I updated the ruby example to include all options in the yaml example. 

I also added a note on configuration processing order.

### DIFF
--- a/bridgetown-website/src/_docs/configuration.md
+++ b/bridgetown-website/src/_docs/configuration.md
@@ -53,9 +53,22 @@ Bridgetown.configure do |config|
   init :dotenv
 
   config.autoload_paths << "jobs"
-
+  url "https://www.bridgetownrb.com"
   permalink "pretty"
   timezone "America/Los_Angeles"
+  template_engine "serbea"
+
+  collections do
+    docs do
+      output true
+      permalink "/:collection/:path.*"
+      name "Documentation"
+    end
+  end
+
+  if Bridgetown.env.development?
+    unpublished true
+  end
 
   only :server do
     init :mail, password: ENV["SENDGRID_API_KEY"]
@@ -64,6 +77,13 @@ end
 ```
 
 The initializer-style config is the most powerful, because you can configure different options for different contexts (static, server, console, rake), as well as interact with environment variables and other system features via full Ruby code. You can also initialize gem-based plugins and configure them in a single pass. And you can write your own initializers which may be called from the main `configure` block.
+
+{%@ Note do %}
+  #### Processing Order
+
+Values from bridgetown.config.yml are processed first, then config/initializers.rb, command line arguments are processed last. If you defined environments in bridgetown.config.yml, but also had a condition in initializers.rb, adding `<pre><%= site.config.to_yaml %></pre>` in a page would show the yaml defined environment, but the active parameters would match those set in initializers.
+{% end %}
+
 
 ## Take a Deep Dive
 
--- a/bridgetown-website/src/_docs/configuration.md
+++ b/bridgetown-website/src/_docs/configuration.md
@@ -81,9 +81,8 @@ The initializer-style config is the most powerful, because you can configure dif
 {%@ Note do %}
   #### Processing Order
 
-Values from bridgetown.config.yml are processed first, then config/initializers.rb, command line arguments are processed last. If you defined environments in bridgetown.config.yml, but also had a condition in initializers.rb, adding `<pre><%= site.config.to_yaml %></pre>` in a page would show the yaml defined environment, but the active parameters would match those set in initializers.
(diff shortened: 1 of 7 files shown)


jaredcwhite:
@brainbuz Fantastic! I'll review soon when I'm able.

brainbuz:
The request in #883 appears to be addressed in the current pagination documentation, this PR shows both ruby and legacy yaml config for enabling pagination. However, I'm finding that following the documentation as written, pagination only works with Liquid templates @jaredcwhite can you take a look?

brainbuz:
Jared I picked this up again and I think I figured out why it's not working. I'll update this PR with any doc fixes later.

brainbuz:
I've experienced the same issue as 'Pages which paginate are excluded from collections.pages' #807. The original requester closed the issue as he decided not to use pagination. In your comments you didn't see an easy fix. What is the best workaround? Could something be done with a partial or component? In my case navigation is using the list of resources in the pages collection to generate a menu. I'd like to request that you reopen 807 until its fixed or a workaround is in the documentation.

jaredcwhite:
Hmm, I'm curious what the error was that prompted your commit [b859405](https://github.com/bridgetownrb/bridgetown/pull/987/commits/b8594051b7b75baa832620b653a6651f872b278f) — the `paginator.each` example should work in ERB, and if not there's a bug somewhere I think.

jaredcwhite:
Oh hey, it just occurred to me that there's now a `original_resource` method on the paginated page, so maybe the docs could mention that rather than just say access to resource info is lost.

brainbuz:
When I tried to paginate with <% paginator.each do |post| %> it did not work,  <% paginator.resources.each do |post| %> did work. original_resource should definitely be in the pagination documentation, as should all of the quirks. 

After creating the single page archives by year alternative, I found I didn't need to use pagination. Most end users don't pay much attention to the urls, so whether a javascript is flipping the page in place or the pages have nice rest urls isn't going to get notice other than by developers.

brainbuz:
If you send me some notes / point at source code or test files with comments etc, I'll be happy to do the writing.

bridgetown-website/src/_docs/configuration.md
```diff
@@ -65,6 +78,13 @@ end
 
 The initializer-style config is the most powerful, because you can configure different options for different contexts (static, server, console, rake), as well as interact with environment variables and other system features via full Ruby code. You can also initialize gem-based plugins and configure them in a single pass. And you can write your own initializers which may be called from the main `configure` block.
 
+{%@ Note do %}
+  #### Processing Order
+
+Values from bridgetown.config.yml are processed first, then config/initializers.rb, command line arguments are processed last. If you defined environments in bridgetown.config.yml, but also had a condition in initializers.rb, adding `<pre><%= site.config.to_yaml %></pre>` in a page would show the yaml defined environment, but the active parameters would match those set in initializers.
```
jaredcwhite, line 84: This isn't strictly true, `site.config` will reflect the most recently-defined config value such as initializers config, not reflect what might have been in the original `bridgetown.config.yml` file. Otherwise, I love this section!

bridgetown-website/src/_docs/configuration/environments.md
```diff
@@ -60,25 +60,24 @@ The default value for `BRIDGETOWN_ENV` is `development`. Thus if you omit
 
 Some elements you might want to hide in development environments include comment forms or analytics. Conversely, you might want to expose an "Edit This Page" button in a development or staging environment but not include it in production environments.
 
-## Environment-specific Configurations
+## Defining Environments
 
-In your `bridgetown.config.yml` config file, as well as the
-`src/_data/site_metadata.yml` metadata file, you can add a block of YAML options
-per environment. For example, given the following metadata:
+The Legacy YAML format defines environments as key/value pairs, config/initilizers.rb uses the environment as a condition for applying settings.
```
jaredcwhite, line 65: spelling issue with "initilizers"

bridgetown-website/src/_docs/content/pagination.md
```diff
@@ -40,7 +40,7 @@ Then you can use the `paginator.resources` logic to iterate through the collecti
 
 {%@ Documentation::Multilang do %}
 ```erb
-<% paginator.each do |post| %>
+<% paginator.resources.each do |post| %>
```
jaredcwhite, line 43: Let's revert this change, I'm convinced we don't need to keep that `resources` method call and if we need to fix a bug to get it working we will (provided I can reproduce).

jaredcwhite (changes requested):
Overall this is looking stellar! I just had a couple of notes to make, should be quick to change and then I'll go ahead and merge. TIA

jaredcwhite:
🎉

PR #988, opened 2025-03-21
fixed frozen string warning for ruby 3.4

surrim:
This is a bug fix.

## Summary

When using rake with Ruby 3.4 I get a warning:

> ~/.gems/gems/bridgetown-foundation-2.0.0.beta4/lib/bridgetown/foundation/questionable_string.rb:8: warning: string returned by :production?.to_s will be frozen in the future

**To reproduce**

```bash
# rbenv install 3.4.2
# rbenv global 3.4.2
bridgetown plugins new XYZ
cd XYZ
rake
```


surrim:
See https://rubyreferences.github.io/rubychanges/3.4.html#warning-to-prepare-for-freezing-string-literals-in-ruby-35

PR #989, opened 2025-03-29
propose cookbook with pagination alternative example

brainbuz:
This is a 🔦 documentation enhancement. 

## Summary

This provides an example of an alternative approach to pagination. 

It proposes a new documentation section: cookbook for code examples.

## Context

The technique could be considered a workaround for #807.

### DIFF
--- a/bridgetown-website/src/_docs/cookbook.md
+++ b/bridgetown-website/src/_docs/cookbook.md
@@ -1,0 +1,21 @@
+---
+order: 2500
+title: Cookbook
+#top_section: Introduction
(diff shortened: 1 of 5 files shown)


jaredcwhite:
I like the Cookbook idea! We had an old site called "Bridgetown Cards" a while back which had something like this going, but since it was a separate site/repo we didn't see much traffic or suggestions. I also tried setting up a similar thing in a forum, but we ended up closing it as the hosting company went haywire on us. Anyway, all that to say: having it in the docs site itself is probably the best option!

PR #998, opened 2025-07-11
Load Dotenv configuration before starting the server to catch additional configuration

### LINKED ISSUE #996: The port number in `puma.rb` config not being honored
If the port number in `config/puma.rb` is anything other than 4000, that's not being honored when Puma starts up…it's still just 4000.

mickeytgl:
This is a 🐛 bug fix.

## Summary

The `BRIDGETOWN_PORT` was not being honored because `Dotenv` hasn't loaded when the `Bridgetown::Commands::Start#start` is being ran and therefore any additional configuration in the `.env` file is being ignored.

This PR requires `dotenv` at the beginning of the file, and loads it before the `ENV["BRIDGETOWN_PORT"]` is fetched. That way, it becomes available and it is taken into account before passing in the configuration to start the serve

## Context

Fixes #996

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/start.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/start.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 require "rackup/server"
+begin
+  require 'dotenv'
+rescue LoadError
+  # Since it's a plugin, we don't want to fail if it's not installed
+end
 
 module Bridgetown
   class Server < Rackup::Server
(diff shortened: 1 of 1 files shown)


jaredcwhite:
Hmm, this is a challenging issue for sure. Ideally, the order of events would be:

* Puma config loads
* dotenv is loaded if need be
* _Then_ the Bridgetown boot process gets the correct port number from the Puma config

I don't think I can merge this as-is since the underlying bug will still remain, aka the port number in the Puma config is essentially meaningless and we might as well remove it from the config entirely. But I'm not sure that feels very idiomatic Puma usage…

jaredcwhite:
Hey @mickeytgl, I have a new PR #1005 to resolve this issue…it was a really tricky problem which I can't even now claim is completely accounted for, so don't feel bad about us not going with this solution. I appreciate you looking into this and your feedback!

mickeytgl:
Thanks @jaredcwhite ! Yeah it's a bit tricker than it seems on principle  😅 Glad you found a solution!

PR #1001, opened 2025-07-27
Handle Symbol#to_s returning chilled strings

michaelherold:
This is a 🐛 bug fix.

## Summary

In [a change to Ruby 3.4][1], the core team made it so `Symbol#to_s` returns a "chilled" string, which is a String that will be frozen in future releases of Ruby.

This change makes two cases that were raising warnings for me on Ruby 3.4.5 use a dup-and-modify approach for handling assignment calls within `method_missing` implementations.

[1]: https://bugs.ruby-lang.org/issues/20350

## Context

See ruby/ruby#12065

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/model/base.rb
+++ b/bridgetown-core/lib/bridgetown-core/model/base.rb
@@ -145,11 +145,7 @@ module Bridgetown
         return attributes[method_name] if attributes.key?(method_name)
 
         key = method_name.to_s
-        if key.end_with?("=")
-          key.chop!
-          attributes[key] = args.first
-          return attributes[key]
-        end
+        return attributes[key.chop] = args.first if key.end_with?("=")
 
         Bridgetown.logger.warn "key `#{method_name}' not found in attributes for " \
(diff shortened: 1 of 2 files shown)


surrim:
Looks very similar to https://github.com/bridgetownrb/bridgetown/pull/988 ✅

michaelherold:
I looked for all calls to `#chop!` so hopefully this is the last of it.

bridgetown-core/lib/bridgetown-core/model/base.rb
```diff
@@ -145,11 +145,7 @@ def method_missing(method_name, *args)
         return attributes[method_name] if attributes.key?(method_name)
 
         key = method_name.to_s
-        if key.end_with?("=")
-          key.chop!
-          attributes[key] = args.first
-          return attributes[key]
-        end
+        return attributes[key.chop] = args.first if key.end_with?("=")
```
jaredcwhite, line 148: Nice! :clap:

jaredcwhite:
Looks great @michaelherold TY!

PR #1007, opened 2025-08-13
Fix routing order dependency in bridgetown_server

michaelherold:
This is a 🐛 bug fix.

- I've added tests

## Summary

The `bridgetown_server` plugin was consuming all routes when `base_path` was empty, preventing other Roda plugins from handling requests. This made plugin order critical and broke existing behavior that allowed plugin usage to be in any order.

This change updates the routing logic to avoid using the terminal `#on(true)` pattern, allowing multiple plugins to coexist regardless of load order.

It also adds regression tests to ensure plugins can handle routes independently.

## Context

#939 was an attempt to reduce duplication in the plugin by relying on the "always" matcher behavior `#on(true)`. This accidentally regressed behavior and made the plugin's method terminal by making the branch _always_ match when `base_path` is the default. In that case, when no sub-branch from either the `ssg` plugin or a `Bridgetown::Rack::Routes` descendant matches, the `#on(true)` form terminates and doesn't pass to the next line in the router.

This change doesn't need a documentation update because the terminal behavior of the plugin was surprising.

### Other possible solutions

I considered removing the fork here and updating the `ssg` plugin and the `Bridgetown::Rack::Routes` logic to include knowledge of `base_path` within them, but I think this solution is cleaner.

- The `ssg` plugin would be straightforward; we could pass the `base_path` in the plugin configuration.
- The `Bridgetown::Rack::Routes` would, I think, continue to have a fork in it because, without the fork, we would end up in the same place due to the reliance on Roda's routing tree logic. Since the fork has to live somewhere, it felt cleaner to have it in the Roda plugin that sets up the other two.

### DIFF
--- a/bridgetown-core/lib/roda/plugins/bridgetown_server.rb
+++ b/bridgetown-core/lib/roda/plugins/bridgetown_server.rb
@@ -157,9 +157,15 @@ class Roda
           scope.initialize_bridgetown_root
 
           base_path = Bridgetown::Current.preloaded_configuration.base_path.delete_prefix("/")
-          on(base_path.empty? ? true : base_path) do
+
+          if base_path.empty?
             ssg # static file server
             Bridgetown::Rack::Routes.load_all scope
+          else
+            on(base_path) do
+              ssg # static file server
+              Bridgetown::Rack::Routes.load_all scope
+            end
           end
         end
       end
--- a/bridgetown-core/test/ssr/server/roda_app.rb
+++ b/bridgetown-core/test/ssr/server/roda_app.rb
@@ -1,7 +1,18 @@
 # frozen_string_literal: true
 
 class RodaApp < Roda
+  module OrderIndependenceExample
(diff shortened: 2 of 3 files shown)


jaredcwhite:
Ah, that makes sense. I have a dim recollection at some point previously I'd thought it isn't ideal that `r.bridgetown` just swallows up everything.

I'm wondering if we can still de-dup via a proc of some kind? Just a suggestion:

```ruby
base_route = proc do
  ssg # static file server
  Bridgetown::Rack::Routes.load_all scope
end

base_path.empty? ? base_route.call : on(base_path, &base_route)
```

michaelherold:
I considered that and felt that it wasn't as clear as the repetition. If we were repeating more, I think it would make more sense to extract the proc.

I guess this gives a spot where more things can go if we need to in the future. 🤔

jaredcwhite:
Sounds good to me. Thanks @michaelherold

PR #1025, opened 2025-09-25
Mostly working docs upgrade from Shoelace to Web Awesome

### LINKED ISSUE #1022: docs: Upgrade from Shoelace to Web Awesome in the site
## Summary

Follow-up to #1018, where Web Awesome was added to bridgetown-core, replacing Shoelace.

The site is still on Shoelace, so this issue is for the site to be updated to use Web Awesome instead of Shoelace.

jaredcwhite:
**Note:** Still a fair bit to do, especially in Dark Mode, but it's a good bunch of effort.

Resolves #1022

### DIFF
--- a/bridgetown-website/frontend/styles/content.css
+++ b/bridgetown-website/frontend/styles/content.css
@@ -120,7 +120,7 @@ article-author {
     width: 100%;
     text-align: center;
 
-    &::part(base) {
+    &::part(body) {
       height: 100%;
     }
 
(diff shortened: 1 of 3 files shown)


jaredcwhite:
This ended up being a whole ton of work that I wasn't fully expecting, and I'm not sure I was happy with the results and haven't felt motivated to go back. I'm now thinking we should stick with Shoelace for now and only switch to Web Awesome when there's a more comprehensive overhaul of the website. Maybe we can use the upcoming Willamette theme as a new base. In any case, I'm just going to close this. I don't feel it's worth the effort right now. 🤷🏻‍♂️

PR #1032, opened 2025-10-06
Fix relative translations for multiple locale files

jclusso:
This is a 🐛 bug fix.

## Summary

Relative translation keys break when used with multiple locale files (those ending in .multi). This PR resolves that by removing the ".multi" left in the key.


jaredcwhite:
Thanks @jclusso, I'm trying to get the GH actions to run but something's wonky. I'll try to run tests locally later on.

I think removing `_` is swapping one bug for another…I'd like to support resources within collections which have root content folders which start with `_`, aka `_docs`, `_episodes`, etc. I'm guessing we need to `delete_prefix` the underscore rather than remove all of them if we want to solve both problems.

jclusso:
> Thanks @jclusso, I'm trying to get the GH actions to run but something's wonky. I'll try to run tests locally later on.
> 
> I think removing `_` is swapping one bug for another…I'd like to support resources within collections which have root content folders which start with `_`, aka `_docs`, `_episodes`, etc. I'm guessing we need to `delete_prefix` the underscore rather than remove all of them if we want to solve both problems.

Ahh, let me try and add a test and fix that handles that aswell.

jclusso:
@jaredcwhite I've pushed another test and fix that should address your concern.

jaredcwhite:
@jclusso That fix looks good. I did track down the CI issue and it's actually an infinite loop. If you cd into bridgetown-core and run `script/test test/test_resource.rb` you'll see that it hangs. I added a tap to the `basename.basename` call and saw that it was getting stuck on a filename of `.nojekyll`. I guess when you have a file which itself starts with a period, you still have that no matter what. Any thoughts on an elegant solution?

jclusso:
@jaredcwhite ahh, didn't realize that was happening. I made a fix. This the best I could think of. Also fixed some other failing tests.

jaredcwhite:
@jclusso thanks, I'll do a final review on this soon.

jaredcwhite:
Looks good! Merging

PR #1040, opened 2025-10-15
Refactor: use Erubi capturi block

### LINKED ISSUE #991: Track: Consider upgrading ERB parser to erubi/capture_block engine
## Summary

Erubi's ERB has added support for capture blocks within `<%=` tags, that is this syntax:

```erb
<%= foobar do %>
  <p>Arbitrary html here</p>
<% end %>
```

Being as this syntax is not supported in native ERB. It is currently implemented via regular expressions in some code borrowed from Rails' implementation of ERB.

See Erubi/CaptureBlockEngine [README](https://github.com/jeremyevans/erubi/tree/master?tab=readme-ov-file#erubicaptureblockengine-) and [implementation](https://github.com/jeremyevans/erubi/blob/master/lib/erubi/capture_block.rb).

## Motivation

It may be a more robust solution to supporting capture blocks than the current regex-based way borrowed from Rails.

## Drawbacks

This would likely only be not implemented if it didn't work. This issue is meant to track the Erubi/CaptureBlockEngine until such time as someone is inspired to try adding it, and it is tested against a test suite of the way people are currently using ERB in Bridgetown.

## Further Context

This proposal to Erubi was introduced by @adam12 in this [Discussion](https://github.com/jeremyevans/erubi/discussions/39). On this page, there is a link to a Gist with a code example and the timestamped link to the (short) talk where this was presented at RubyWorld conference. The talk is in Japanese but dubbed over by a robot in English. The novel idea and first implementation was given by Akira Matsuda, who gives the talk in the link. The Erubi implementation was implemented by @adam12 and @jeremyevans.

fpsvogel:

## Summary

Updates ERB parsing to use Erubi's capture blocks instead of the regex-based approach borrowed from Rails.

It may be easiest to review this PR's commits individually.

## Context

Resolves #991

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
+++ b/bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
@@ -24,6 +24,10 @@ module Bridgetown
       @buffer.encode!
     end
 
+    def initialize_copy(other)
+      @buffer = other.to_str
+    end
+
     # Concatenation for <%= %> expressions, whose output is escaped.
     def <<(value)
       return self if value.nil?
--- a/bridgetown-core/test/test_output_buffer.rb
+++ b/bridgetown-core/test/test_output_buffer.rb
@@ -2,14 +2,13 @@
 
 require "helper"
 
-module Bridgetown
-  class OutputBufferTest < BridgetownUnitTest
-    def setup
-      super
+class TestOutputBuffer < BridgetownUnitTest
+  describe "output buffer" do
+    before do
       @buffer = Bridgetown::OutputBuffer.new
     end
 
-    should "be able to be duped" do
+    it "should be able to be duped" do
       @buffer << "Hello"
       copy = @buffer.dup
       copy << " world!"
@@ -18,8 +17,8 @@ module Bridgetown
       assert_equal "Hello world!", copy.to_s
     end
 
-    context "#<<" do
-      should "maintain HTML safety" do
+    describe "#<<" do
+      it "maintains HTML safety" do
         @buffer << "<p>Nothing bad to see here.</p>"
 
         assert_predicate @buffer, :html_safe?
@@ -28,8 +27,8 @@ module Bridgetown
       end
     end
 
-    context "#safe_append=" do
-      should "bypass HTML safety" do
+    describe "#safe_append=" do
+      it "bypasses HTML safety" do
         @buffer.safe_append = "<p>Nothing bad to see here.</p>"
 
         assert_predicate @buffer, :html_safe?


bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
```diff
@@ -15,30 +15,15 @@ def <<(value)
...
   end
 
   class ERBEngine < Erubi::CaptureBlockEngine
     private
 
-    def add_code(code)
-      @src << code
-      @src << ";#{@bufvar};" if code.strip.split(".").first == "end"
-      @src << ";" unless code[Erubi::RANGE_LAST] == "\n"
-    end
```
fpsvogel: I'm removing this override because it *seems* to be specific to blocks, but I'm not entirely sure because if I remove this method even on `main`, the tests still pass.

The original method in `Erubi::Engine`: https://github.com/jeremyevans/erubi/blob/f3c93cc46aa09396dafa9c422e06ec6ad22e0db2/lib/erubi.rb#L231-L237

bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
```diff
@@ -41,23 +42,6 @@ def add_text(text)
...
-                       else
-                         ".append="
-                       end
-
-      if BLOCK_EXPR.match?(code)
-        src << " " << code
-      else
-        src << "(" << code << ");"
-      end
-    end
```
fpsvogel: The original method in `Erubi::Engine`: https://github.com/jeremyevans/erubi/blob/f3c93cc46aa09396dafa9c422e06ec6ad22e0db2/lib/erubi.rb#L239-L247

bridgetown-core/lib/bridgetown-core/converters/erb_templates.rb
```diff
@@ -40,16 +31,10 @@ def |(value)
...
-
     def add_text(text)
       return if text.empty?
 
-      src << bufvar << ".safe_append='"
+      src << bufvar << ".safe_concat'"
       src << text.gsub(%r{['\\]}, '\\\\\&')
       src << "'.freeze;"
     end
```
fpsvogel: This override needs to stay, because the original method in `Erubi::Engine` generates a call to the buffer's `<<`, which escapes content. And we don't want text to be escaped.

The original: https://github.com/jeremyevans/erubi/blob/f3c93cc46aa09396dafa9c422e06ec6ad22e0db2/lib/erubi.rb#L220-L229

bridgetown-core/test/test_erb.rb
```diff
@@ -35,6 +35,10 @@ def setup
     should "properly handle block expressions" do
       assert_includes @erb_page.output, "\n===\n+Value: value+\n---\n"
     end
+
+    should "not escape expressions in <%== %>" do
+      assert_includes @erb_page.output, "<em>This is an unescaped expression & it shouldn't be escaped</em>"
+    end
```
fpsvogel: This isn't a new feature, just a test case that was missing. (It passes in `main` too.)

jaredcwhite:
This looks great @fpsvogel! I want to poke it a bit in some local testing, but just reviewing the changes I think we should be all set with this.

jaredcwhite:
FYI I'm doing a three-way merge and including a couple changes related to pulling #723 in as well. Might as well!

PR #1046, opened 2025-10-21
Support pure Ruby (`.rb`) layouts

### LINKED ISSUE #1038: Pure Ruby templates (.rb) aren't supported as layouts
This is a bit esoteric, but it seems that while we support resource templates as pure Ruby files, we don't support layouts…a limitation of the `ruby_templates.rb` converter.

jaredcwhite:
* also fix some issues with front matter line counts which affected exception backtraces
* also use correct front matter line count for layouts

Resolves #1038 

BTW I'm really digging how cool the Streamlined syntax is for this. Check it out!

```ruby
###ruby
front_matter do
  layout :default
  custom_var 123
end
###

render html->{ <<~HTML
  <h1>#{text->{ resource.data.title }}</h1>
HTML
}

render { yield }

render html->{ <<~HTML
  <aside>Custom var: #{text->{ layout.data.custom_var }}</aside>
HTML
}
```

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/commands/new.rb
+++ b/bridgetown-core/lib/bridgetown-core/commands/new.rb
@@ -99,7 +99,7 @@ module Bridgetown
         "esbuild"
       end
 
-      def postcss_option
+      def postcss_option # rubocop:disable Naming/PredicateMethod
         !options["use-sass"]
       end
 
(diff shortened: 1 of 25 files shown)


fpsvogel (commented):
I'm going to look at this PR more carefully later today to learn how it works, but Q: in principle, could this also work for pure Ruby components? And is that a thing that we want / should allow?

jaredcwhite:
@fpsvogel That's a feature we have already…as long as you define a `template` method and return a string or string-like object, you're good to go. And we have the new Streamlined syntax to make that easier:

https://www.bridgetownrb.com/docs/components/ruby#use-bridgetowncomponent-for-template-rendering

https://www.bridgetownrb.com/docs/template-engines/erb-and-beyond#streamlined

fpsvogel:
@jaredcwhite Oh cool, the Streamlined syntax is what I was thinking. I missed it back in 2.0!

PR #1053, opened 2025-10-27
Remove Active Support as a dependency

### LINKED ISSUE #1016: Complete the removal of Active Support dependency
Continuing on work done in #881, this is just a stub until I can get a full sense of what the final set of changes will be which I will document here or in the PR.

Working through the test suite and fixing things thus far, I'm at about 20 files changed and there is only one error remaining, which is rad!

jaredcwhite:
Resolves #1016

(note: there may be a few additional things we want to do with String refinements)

- [x] Tests
- [ ] Documentation (mentioning that people can add it back to their Gemfile if they need some functionality)


jaredcwhite:

PR #1055, opened 2025-10-31
Add `ruby_render` Liquid tag

### LINKED ISSUE #976: feat: erb_render method to match liquid_render
## Summary

The liquid_render method allows a ruby template to call a liquid component. It would make sense to be able to call a ruby component from a liquid template.

## Motivation

It just makes sense in light of the philosophy of being able choose and switch freely between template engines. 

https://edge.bridgetownrb.com/docs/template-engines contains: You can mix ‘n’ match template types in the same project. For example, Liquid’s simple syntax and safe execution context make it ideal for designer-led template creation, so you could use Liquid for layouts but stick to ERB for code-intensive pages and other resources.

Which implies that you can do this.

If the feature already exists, it needs to be documented.

fpsvogel:

## Summary

Adds a Liquid tag `ruby_render` that renders a Ruby expression or a simple Ruby component. "Simple" meaning without content in a block:

```
{% ruby_render MyComponent.new(title: "My Title") %}
```

## Context

Resolves #976 

## Manual test

I created a new site with `bridgetown new test_site -t liquid`, then created two Ruby components:

- one copied from [bridgetown-core/test/source/src/_components/ruby_component.rb](https://github.com/bridgetownrb/bridgetown/blob/c3d1d63d2318b8f440fcdd3634e67f8889eee660/bridgetown-core/test/source/src/_components/ruby_component.rb)
- the other a component with an ERB template

Then I rendered these two components plus a Ruby expression (to show that's possible too) at the top of the blog post:

```
{% ruby_render RubyComponent.new %}
{% ruby_render TemplatedComponent.new(name: "Felipe") %}
{% ruby_render "A Ruby string with the random number #{rand(100)}" %}
```

image

## Added section on website "Ruby Components" page

Adapted from [the section on liquid_render](https://www.bridgetownrb.com/docs/components/liquid#rendering-liquid-components-from-ruby-based-templates)

image

### DIFF
--- a/bridgetown-website/src/_docs/components/ruby.md
+++ b/bridgetown-website/src/_docs/components/ruby.md
@@ -271,10 +271,10 @@ In addition to rendering a template for you, `Bridgetown::Component` provides a 
 You can use the `ruby_render` helper from Liquid templates to render Ruby components.
 
 ```liquid
-{% ruby_render "test_component", title: "Ruby FTW!", my_values: [20, 30] %}
+{% ruby_render "test_component", title: "Ruby FTW!" %}
 ```
 
-In this example, `TestComponent` is initialized with the keyword arguments `title` and `my_values`, then rendered.
+In this example, `TestComponent` is initialized with the keyword argument `title`, then rendered.
 
 [Passing in content via a block](#content) is not yet supported.
 


bridgetown-core/test/test_tags.rb
```diff
@@ -3,20 +3,34 @@
 require "helper"
 
 class TestTags < BridgetownUnitTest
+  using HashWithDotAccess::Refinements
```
fpsvogel: I tried nested Structs too, instead of `.as_dots` [below](https://github.com/bridgetownrb/bridgetown/pull/1055/files#diff-b019be4d9e9197f56f495d2d206130315a63806b1916683300a4bc051bc5fe1bR31), but that didn't make the tests in this file run noticeably faster.

bridgetown-core/test/test_tags.rb
```diff
@@ -3,20 +3,34 @@
...
+
   include ActiveSupport::Testing::TimeHelpers
 
   def setup
     FileUtils.mkdir_p("tmp")
   end
 
-  def create_post(content, override = {}, converter_class = Bridgetown::Converters::Markdown) # rubocop:disable Metrics/AbcSize
+  def create_post(content:, override: {}, converter_class: Bridgetown::Converters::Markdown, page_title: nil) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
fpsvogel: I made these into keyword arguments in [a separate commit](https://github.com/bridgetownrb/bridgetown/pull/1055/commits/8defa340c01581b075f14187c5dcd817be2396b8) to avoid [one gross call to this method](https://github.com/bridgetownrb/bridgetown/pull/1055/commits/8defa340c01581b075f14187c5dcd817be2396b8#diff-b019be4d9e9197f56f495d2d206130315a63806b1916683300a4bc051bc5fe1bL110). If it's preferable to keep the positional args, I'm fine omitting that commit.

bridgetown-core/lib/bridgetown-core/tags/ruby_render.rb
```diff
@@ -0,0 +1,30 @@
...
+      def initialize(tag_name, ruby_expression, tokens)
+        super
+        @ruby_expression = ruby_expression
+      end
+
+      # @param context [Liquid::Context]
+      # @return [String]
+      def render(context)
+        view_context = Bridgetown::RubyTemplateView.new(context.registers[:resource])
+        result = eval(@ruby_expression) # rubocop:disable Security/Eval
```
fpsvogel: If we want to avoid `eval`, it might be better to use a non-Ruby syntax for the Liquid tag contents (such as [the one we previously discussed](https://github.com/bridgetownrb/bridgetown/issues/976#issuecomment-3463297139) in the GH Issue), because the other way to parse the contents (apart from `eval`) is via regex, and trying to emulate Ruby parsing via regex seems difficult to do well, even if we allow nothing other than component instantiation.

If the regex route is preferable, I assume it's because `eval` is unsafe, but I would be curious to know how specifically, for this use case.

bridgetown-website/src/_docs/components/ruby.md
```diff
@@ -266,6 +266,16 @@ In addition to rendering a template for you, `Bridgetown::Component` provides a
 * `render?` – if you define this method and return `false`, the component will not get rendered at all.
 * `before_render` – called right before the component is rendered when the view_context is known and all helpers available.
 
+## Rendering Ruby Components from Liquid Templates
+
+You can use the `ruby_render` helper from Liquid templates to render Ruby components.
```
fpsvogel: Ruby components are the primary use case, so I'm not sure if it's worth mentioning that it's possible to render any Ruby expression—if it's not a component (i.e. doesn't have a `:render_in` method) then it's rendered with #to_s ([see above](https://github.com/bridgetownrb/bridgetown/pull/1055/files#diff-a47f8b064c8def2446488a39a9de157f781a29ced7e915964afb69657132da65R23)).

jaredcwhite:
Some cool stuff in here @fpsvogel. I'll want to think on this a bit…one of the promises of Liquid is that you're working within a "sandboxed" environment and all the values/tags/etc. live with the Liquid environment. If we punch a hole right into Ruby land with `eval`, it could end up with unexpected weird behavior. I'm not worried about the security aspect per se as Bridgetown isn't built around a "user-contributed" Liquid template content model, and there are no guarantees of safety in that respect, but there could be other gotchas.

fpsvogel:
@jaredcwhite Any further thoughts on this now that we're looking ahead to 2.2?

If we decide against allowing Ruby syntax, another possibility that we talked about in the issue is to use Liquid syntax:

```liquid
{% ruby_render "my_component", title: "My Title" %}
```

jaredcwhite:
@fpsvogel I think it would be good to try to stick with Liquid syntax, maybe we can look to one of the other tags like `class_map` to see how those arguments are parsed.

fpsvogel:
@jaredcwhite I pushed up one approach to Liquid syntax.

This approach [still leans on eval](https://github.com/bridgetownrb/bridgetown/pull/1055/changes#diff-a47f8b064c8def2446488a39a9de157f781a29ced7e915964afb69657132da65R26-R27) quite a bit. Take this example:

```liquid
{% ruby_render "card", title: "Hello", footer: "I am a card" %}
```

Everything after `"card",` is eval'ed as the arguments to `Card.new(...)`.

EDIT: Or, more to my point below, an example with non-string args:

```liquid
{% ruby_render "number_cruncher", crunchiness: 11, numbers: [1, 6, 8] %}
```

I thought about following the pattern in the `class_map` tag: iterating over the comma-separated items to build the arguments to `Card::new`. But even then we'd still use `eval` for each value, because unlike in `class_map`, the values can be not just strings but potentially numbers, symbols, arrays, etc. I'm not sure what would be gained by restricting `eval` to each value, rather than the whole string of arguments (as I've done).

jaredcwhite:
Sorry for the wait @fpsvogel …hoping to get a change to look at this soon!

jaredcwhite:
@fpsvogel So one of the issues with the `eval` approach is it breaks the encapsulation of the Liquid template. For example, if you'd created a variable in Liquid:

```liquid
{% assign show_border = false %}
```

and then used this tag:

```liquid
{% ruby_render "thumbnail", border: show_border %}
```

it would crash because there's no `show_border` within the eval context.

This was a thorny issue and I didn't fully understand the best way to remedy, so I did some digging around in Liquid's source code and eventually arrived at what seems to be a pretty clean solution. I took the liberty of committing a fix to your branch, feel free to tweak it to your liking. Let me know if that makes sense to you!

fpsvogel:
@jaredcwhite Brilliant! That is a problem I hadn't thought about, and your solution looks very clean to me.

jaredcwhite:
Anything else you think this PR needs? Or is it ready for final review and merge?

fpsvogel:
@jaredcwhite Just now I made [a change](https://github.com/bridgetownrb/bridgetown/pull/1055/commits/b226e36449584d8dc3356d0e5c267a63327e8c16) omitting an array literal from the `ruby_render` docs, because Liquid does not support array literals.

I think this PR is ready now.

PR #1057, opened 2025-11-14
Bump minimum Ruby version to 3.2, Node version to 22

### LINKED ISSUE #1045: Bump minimum Ruby version to 3.2, Node version to 22
## Summary

It looks like it's time to bump the minimum Ruby version to 3.2:

- Bridgetown aims to support the past two Ruby versions before the current (according to [the most recent PR to bump the Ruby version](https://github.com/bridgetownrb/bridgetown/pull/841)). That would be 3.2, 3.3, and the current 3.4.
- Ruby 3.1 [is EOL](https://www.ruby-lang.org/en/downloads/branches/).

EDIT: also time to bump Node to version 22 (the next LTS version), see comment below.

fpsvogel:

## Summary

Bumps Ruby and Node versions.

Includes adjustments for Ruby changes. The commit messages include links to specific points in the Ruby changelogs.

It'll make the most sense to review this PR one commit at a time.

## Context

Resolves #1045

### DIFF
--- a/bridgetown-website/src/_data/requirements.yml
+++ b/bridgetown-website/src/_data/requirements.yml
@@ -1,2 +1,2 @@
 min_ruby: 3.2
-min_node: 20.6
+min_node: 22.21


fpsvogel:
@jaredcwhite Where does the Node version need to be bumped?

I found `bridgetown-website/src/_data/requirements.yml`, but where else besides that? That's the only place where I see an occurrence of the previous version (20.6).

jaredcwhite:
@fpsvogel I think maybe we don't actually manage it anywhere, it's just documented. I use nvm on my own machines with a `.nvmrc` file, but Bridgetown doesn't generate that by default.

jaredcwhite:

Issue #1074, opened 2026-01-16
External source include _ pages in their collection


mperham:
I have a set of Github wiki pages pulled in as external content. One of the elements in the wiki is `_Footer.md` which github will display at the bottom of each page. This results in a malformed page in the collection. For instance, when I generate my sitemap, I get this entry:

```
        <url>
          <loc>http://localhost:4000/wiki/.</loc>

              <xhtml:link rel="alternate" hreflang="x-default" href="http://localhost:4000/wiki/." />
              <xhtml:link rel="alternate" hreflang="en" href="http://localhost:4000/wiki/." />

          <lastmod>2026-01-16T11:19:46-08:00</lastmod>


        </url>
```

I would expect the external collection not to include underscored records.

fpsvogel:
@mperham The next release will include filtering of external content. To omit files that start with an underscore such as `_Footer.md`, you'll want to make this addition to your `initializers.rb`:

```ruby
  init :external_sources do
    contents do
      wiki "wiki"
    end
    
    # this part is new
    filters do
      wiki ->(name, path) { !name.start_with?("_") }
    end
  end
```

PR #1078, opened 2026-02-04
Add hook to filter external source files

fpsvogel:

## Summary

Adds a config option to filter files from external content sources based on file name.

## Context

Resolves https://github.com/bridgetownrb/bridgetown/issues/1074

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/readers/plugin_content_reader.rb
+++ b/bridgetown-core/lib/bridgetown-core/readers/plugin_content_reader.rb
@@ -23,7 +23,7 @@ module Bridgetown
       end
     end
 
-    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     def read_content_root(collection_name, content_dir)
       collection = site.collections[collection_name]
       unless collection
@@ -35,17 +35,16 @@ module Bridgetown
       end
 
       Find.find(content_dir) do |path|
-        if site.config.external_sources_filters &&
-            site.config.external_sources_filters[collection_name] &&
-            !site.config.external_sources_filters[collection_name].call(File.basename(path), path)
-          Find.prune if File.directory?(path)
+        if File.directory?(path)
+          if rejected_by_external_sources_filter?(path, collection_name) ||
+              EntryFilter::SPECIAL_LEADING_CHAR_REGEX.match?(File.basename(path))
+            Find.prune
+          end
+
           next
         end
 
-        if File.directory?(path)
-          Find.prune if EntryFilter::SPECIAL_LEADING_CHAR_REGEX.match?(File.basename(path))
-          next
-        end
+        next if rejected_by_external_sources_filter?(path, collection_name)
 
         if File.symlink?(path)
           Bridgetown.logger.warn "Plugin content reader:", "Ignored symlinked asset: #{path}"
@@ -54,7 +53,14 @@ module Bridgetown
         end
       end
     end
-    # rubocop:enable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
+    # rubocop:enable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+    def rejected_by_external_sources_filter?(path, collection_name)
+      filter = site.config.external_sources_filters&.dig(collection_name)
+      return false unless filter
+
+      !filter.call(File.basename(path), path)
+    end
 
     # @param collection [Bridgetown::Collection]
     def read_content_file(content_dir, path, collection)


bridgetown-core/test/features/test_external_sources.rb
```diff
@@ -15,6 +15,8 @@ class TestExternalSources < BridgetownFeatureTest
             contents do
               pages "#{sources_folder}"
             end
+
+            config.external_sources_filter = ->(basename, path) { !basename.start_with?("_") }
```
fpsvogel: I wanted to make the filter defined via a DSL, like this:

```ruby
filter do |basename, path|
  !basename.start_with?("_")
end
```

But after poking around in the configuration DSL for a while, I couldn't figure out how to achieve this.

bridgetown-core/lib/bridgetown-core/readers/plugin_content_reader.rb
```diff
@@ -23,6 +23,7 @@ def read
       end
     end
 
+    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
```
fpsvogel: Once I get a definite sense of direction on this PR, I'll refactor this method to avoid these RuboCop exceptions.

fpsvogel: Actually… how strongly should we try to adhere to RuboCop? In this case, splitting up the method might not actually make it easier to understand.

bridgetown-core/test/features/test_external_sources.rb
jaredcwhite: @fpsvogel That API looks pretty slick! I'm not sure about doing away with the full path though…what if someone wants to filter out an entire folder or certain files if they're in certain folders, etc.?

fpsvogel: Good point, I hadn't thought about the scenario of filtering out certain files if they're in certain folders.

I added back the path and adjusted the docs in a new commit, https://github.com/bridgetownrb/bridgetown/pull/1078/changes/80036f0a941923d0071c0cfa606a31fdd87384f6

bridgetown-core/lib/bridgetown-core/readers/plugin_content_reader.rb
jaredcwhite: I'm pretty liberal with disabling cops if I don't feel it's helpful in context…we could also look into loosening some of these metrics if it's just too annoying too often.

fpsvogel:
@jaredcwhite We're pruning directories that don't match the filter, i.e. filtering out everything within a directory.

I realized the tests don't cover that, so I added an expectation in https://github.com/bridgetownrb/bridgetown/pull/1078/changes/3582a169ca2f5ffd28b0637d1597ec13a8f832fe. Then I improved the test in https://github.com/bridgetownrb/bridgetown/pull/1078/changes/442e7a2a038f6e16b590359d685ee1c11bd2f11f) so that `EntryFilter::SPECIAL_LEADING_CHAR_REGEX` doesn't cause the directory starting with `_` to be pruned ([here](https://github.com/bridgetownrb/bridgetown/pull/1078/changes#diff-59bbf65087c1f1a53c7e7d53812b3fe400166455a7fc6dbb94f39038f3dad607R46)) even if the filter-related prune is removed.

So now if you comment out the filter-related directory prune ([here](https://github.com/bridgetownrb/bridgetown/pull/1078/changes#diff-59bbf65087c1f1a53c7e7d53812b3fe400166455a7fc6dbb94f39038f3dad607R41)), the test will fail at [this assertion](https://github.com/bridgetownrb/bridgetown/pull/1078/changes#diff-dc863dcb07b3c030e458b98b89d6289823a649d66459a7e3572f2474b0e2da09R51) because the file within that directory is included despite the directory not matching the filter.

jaredcwhite:
@fpsvogel Thanks for continuing to crank on this! So I think my original point still stands, which is the following scenario:

* I would like to filter out any files starting with "skip-me"
* I add the filter `!name.start_with?("skip-me")`
* Now in a folder of 10 files, only 8 files get processed because the other 2 files are `skip-me-abc.md` and `skip-me-xyz.md`

So in that scenario, we wouldn't want to skip over that entire folder, just those 2 individual files within the folder. Does make sense?

fpsvogel:
@jaredcwhite I updated the test so that I think it covers your scenario:

With a filter of `!name.start_with?("skipme_")`:

- `external_sources_folder/subfolder/skipme_notes.md` is omitted
- `external_sources_folder/subfolder/external_page.erb` is included
- `external_sources_folder/skipme_subfolder` and its contents are omitted
  - This is the only one where `Find.prune` is executed; without `Find.prune`, `external_sources_folder/skipme_subfolder/notes.md` would still be included.

Let me know if I'm still missing something!

jaredcwhite:
I think we may be getting a false positive on the mixed case, because the pruning is still happening at the folder level.

The `bridgetown-core/test/external_sources_folder/subfolder/skipme_notes.md` file is getting skipped over, yes, but I suspect if you added another file to that `subfolder` that doesn't start with `skipme`, it wouldn't get processed, because that entire subfolder is being pruned. We'd want everything in that subfolder, *minus* the `skipme` file(s).

jaredcwhite:
@fpsvogel Hey, sorry for being so pedantic…I'm realizing I missed understanding what the `next` line did in the conditional you added, so I think this looks great. I appreciate you adding the extra tests just to be sure. As far as I'm concerned, LGTM!

fpsvogel:
@jaredcwhite Yep no worries, I was writing up a reply when I noticed the logic there was hard to follow. So I restructured it to what I hope is an easier-to-read form.

If that adjustment looks OK to you, I'll go ahead and merge.

jaredcwhite:
Clean. I like it!

Issue #950, opened 2024-12-04
Simplify live reload SSE endpoint after Puma bug is fixed


ayushn21:
While pairing with @ioquatix, we looked at the live reload code in Bridgetown to try and make it work with both Falcon and Puma.

Currently we're spawning a background thread to hold the connection, which is at best a hack, and at worst, wrong. We need to do this because Ctrl+C doesn't kill Puma without it. This turns out to be a bug in Puma itself. We've filed an issue here: https://github.com/puma/puma/issues/3569.

After this is fixed, we should be able to remove the background thread from the SSE endpoint without any issues.

jaredcwhite:
We can close this now, yes?

ayushn21:
Yep! I thought merging the PR would've automatically closed it :)

PR #1083, opened 2026-02-27
Swap Puma out in favour of Falcon

### LINKED ISSUE #951: Swap Puma in favour of Falcon as the default web server
Here's my pitch to swap our default web server. I know we're still only getting v2 out the door, but this is one I'd like to do for v3 ... so a longer term thing.

## Why we should

- Falcon is seriously high performance. Puma is bound by threads, we set a number when we boot Puma up and it's usually in the 10s. Falcon uses fibers .... we don't configure these ... for all practical purposes there isn't an upper limit (technically I believe there is, but we're talking 10s of thousands).
- Falcon is HTTP/2 native.
- Falcon is built with streaming in mind. Handling a websocket or SSE connection with Puma is not for the faint hearted. It ties up a thread.... the developer really needs to think about it properly. Action Cable does really funky socket hijacking to take the connection out of Puma. With Falcon, you don't need to worry about that any more ... the way it's built, these things "just work".
- We can use Falcon for SSL termination ... there's no need to have Caddy/Nginx or something like that in front of it. This will seriously simplify deployments.
- Falcon is pure Ruby. We should encourage other projects in the Ruby ecosystem, and this is a great chance to have the entire Bridgetown infrastructure run on Ruby without any other pieces like Caddy etc. 

## Potential pitfalls

- Falcon is still very new and adoption is limited. But we can and should encourage more widespread use.
- It's not as battle tested as incumbents so there could be security issues.
- The community is still fairly small.

## Integration issues

I honestly don't think there should be any. The live reload code is the only holdup at the moment and that's easy enough to sort out if we swap to puma. There's a couple of more pieces we need to sort out for Rack compliance, but it's not too relevant I don't think. It should be a straight swap for the most part.

Alongside Falcon, we can also introduce our opinionated deployment tool which I'm currently working on. I've been working with Caddy for SSL termination and a reverse proxy, but bringing all that under the same roof under Falcon would just simplify the setup so much. And it'd all be Ruby!

ayushn21:
An initial stab at swapping Puma out for Falcon.

At first look, everything seems to be working. We removed all hard dependencies on Puma last year so we might be mostly good to go here. @jaredcwhite @fpsvogel could you please give it a try on some of your websites and see how you get on? Make the following changes to your `Gemfile`:

```diff
-gem "puma"

+gem "bridgetown-core", github: "ayushn21/bridgetown", branch: "falcon"
+gem "falcon"
```

Then run `bundle` and you should be able to start your server with `bin/bt start`. 

## Further steps / unknowns

- [x] Remove references to Puma in code comments.
- [x] Update site template to use Falcon.
- [x] Swap out `puma.rb` for `falcon.rb` (if needed).
- [x] Update docs as necessary.

Resolves [#4567](https://github.com/bridgetownrb/bridgetown/issues/951), https://github.com/bridgetownrb/bridgetown/issues/950

## Notes

> Can we run HTTPS locally? `falcon serve` serves the website over HTTPS, but doesn't run the Bridgetown build so can't be used in development.

We can't with `Rackup`. In a subsequent PR, I'll investigate a way to invoke the web server directly in the `start` command alongside the Bridgetown build. This will benefit Puma users as well because `bundle exec puma` offers more flexibility than Rackup.


ayushn21:
Hey @ioquatix, when you have a moment I'd really appreciate some guidance with running HTTPS locally via our `bin/bt start` command.

We have a `Bridgetown::Server` class which subclasses `Rackup::Server` and that's what we're starting up with `bin/bt start`. 

```ruby
Bridgetown::Server.new({
  Host: bt_options.bind,
  Port: port,
  config: rack_config_file
}).tap do |server|
  if server.serveable?
    # ...
  
    server.start(after_stop_callback)
  else
    say "Unable to find a Rack server."
  end
end
```

Is there a way to setup SSL here in a generic way that wouldn't break if the user was using another server like Puma, or is it too much of a hassle best left as is?

Thanks :D

ayushn21:
After some further research, I don't think SSL via the `Rackup::Server` class is possible, or a good idea. The Falcon docs say:

```
You can invoke Falcon via rackup:

$ rackup --server falcon
This will run a single-threaded instance of Falcon using http/1. While it works fine, it's not recommended to use rackup with falcon, because performance will be limited.
```

Perhaps there's another approach to explore, I'll have a think.

ioquatix:
Rackup needs an overhaul. Other than that, the best advice is just to use `config.ru` and advise users to pick a server.

Regarding SSL, falcon has great support for it, using the `localhost` gem for self-signed local TLS, it mostly just works out of the box. https://socketry.github.io/localhost/guides/getting-started/index

I wish there was a better answer.

ayushn21:
Thanks so much @ioquatix  ... thought that might be the case :).

jaredcwhite:
I think saying we now offer "beta" status for folks to try out Falcon is a good idea, but I'm very hesitant to change any behavior for existing & new sites which are using Puma. I'm wondering how we could modify this PR so essentially nothing has changed in the context of using Puma, but still also support the Falcon use case.

ayushn21:
> nothing has changed in the context of using Puma, but still also support the Falcon use case.

Nothing will change for existing sites. Unless you explicitly add Falcon to your `Gemfile` and remove Puma, your site will keep working as it was. 

Since we removed our hard dependency on Puma last year, this was actually a pretty straightforward change. All new sites will be created with `falcon`, but I'll add a `--puma` option or something like that to use Puma.

Nothing changes for existing sites unless you manually switch over to Falcon.

ayushn21:
@jaredcwhite I think this is pretty much good to go now. New websites will have falcon installed by default. You can use `bridgetown new my_site -s puma` to install puma instead.

Nothing changes for existing site, they should still work as is and folks can swap in Falcon or whatever other Rack server as they wish.

I've updated the docs to remove all mentions of Puma. 

Next step is to refactor the `start` command. Rackup is soft deprected and the Rack core team's recommendation is to use the server's command line program to start the server. As such this command will need refactoring to remove the dependency on `Rackup::Server`. I have a few ideas, will experiment and make a new PR.

bridgetown-website/src/_docs/configuration/web-server.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Web Server Configuration
+order: 0
+top_section: Configuration
+category: customize-your-site
+back_to: configuration
+---
+
+Bridgetown requires a [Rack-compliant](https://rack.github.io/rack) web server, ([along with Roda](/docs/routes)) for serving up statically-built files as well as any dynamic routes. The default server is [Falcon](https://socketry.github.io/falcon), but we also natively support [Puma](https://puma.io).
```
ayushn21: I've renamed this page to offer more generic guidance since the user can bring whichever rack server they wish.

Also, I've removed some of the more specific configuration options. The web server's docs should be used for this.

Longer term, once we drop  the `Rackup` dependency, the user will have to configure the server directly anyway.

jaredcwhite:
@ayushn21 I'll do an official review shortly (mainly just a couple of questions), but I gave this a peek again and it looks awesome! Running it on a test site and all seems good. Glad you got the live reload errors thing figured out.

ioquatix (approved):

bridgetown-core/lib/bridgetown-core/commands/build.rb
```diff
@@ -87,7 +87,7 @@ def watch_site(config_options) # rubocop:disable Metrics
           external_ip = Socket.ip_address_list.find do |ai|
             ai.ipv4? && !ai.ipv4_loopback?
           end&.ip_address
-          scheme = config_options.bind&.split("://")&.first == "ssl" ? "https" : "http"
+          scheme = config_options.url&.split("://")&.first == "ssl" ? "https" : "http"
```
jaredcwhite: So I think `bind` in this case was a sort of Puma-ism…we'd never have a regular URL starting with `ssl`, so this code may not be quite right.

bridgetown-website/src/_docs/configuration/web-server.md
```diff
@@ -0,0 +1,28 @@
...
+
+In development (`bin/bridgetown start`), Bridgetown uses [Rackup](https://github.com/rackup) to start the web server. Since this is a generic interface, the configuration options are limited. By default, Bridgetown serves over HTTP via port `4000`, bound to `0.0.0.0` (this makes it accessible via localhost as well as the network).
+
+You can change the port using `-p`:
```
jaredcwhite: This should be uppercase `-P`

bridgetown-website/src/_docs/configuration/web-server.md
```diff
@@ -0,0 +1,28 @@
...
+
+```shell
+$ bin/bridgetown start -p 6000
+```
+
+And, to change the IP address to something other than `0.0.0.0`, provide a `--bind` / `-B` argument:
+
+```shell
+$ bin/bridgetown start -b 192.168.1.1
```
jaredcwhite: This should be uppercase `-B`

jaredcwhite (approved):
Just found a couple of docs nits, and there's one thing about the config I'm puzzled by, other than those comments, this looks awesome to me. Very cool stuff!

bridgetown-core/lib/bridgetown-core/commands/build.rb
ayushn21: Ah ok, I can't remember why I made this change haha, been too long but must've been that something didn't quite work in my testing. 

Either way, https://github.com/bridgetownrb/bridgetown/pull/1093 deletes this code so let's not touch it until that's merged. Will revert this line and merge this PR :)

bridgetown-website/src/_docs/configuration/web-server.md
ayushn21: The number of times I've made this mistake, you'd think I'd have learned my lesson by now 🤣

PR #1088, opened 2026-04-02
Drop Ruby 3.2 support and add Ruby 4

ayushn21:
Ruby 3.2 is now EOL: https://www.ruby-lang.org/en/downloads/branches/#ruby-32

Dropping 3.2 and adding 4.0 to our test matrix. Also updated the Gemspec with 3.3.0 as the minimum Ruby version.


jaredcwhite:
Awesome sauce! I need to start planning the v2.2 release, so this would be a good opportunity to push forward. Maybe we also move Node minimum to v23 or something (less of an imperative though).

jaredcwhite:
Looked into it…I think we can stick with Node 22 as minimum for now.

Oh I just realized this didn't update the min requirements in `src/_data/requirements.yml`. I can go take care of that.

PR #1091, opened 2026-04-07
Fix typo in rdoc comment

ayushn21:
As it says on the tin ...


ayushn21:
Merging because it just removes one word from a comment ...

PR #1092, opened 2026-04-15
Optimize all_locales with pre-computed locale index

jclusso:
## Summary

`Localizable#all_locales` currently performs a linear scan of all resources in the collection on every call. For sites with many localized pages, this becomes O(n²) — each of n resources scans all n resources to find its locale variants.

This PR builds a locale index once before rendering that groups resources by their locale-independent identity (collection + slug + localeless path). This replaces the O(n) per-call scan with an O(1) hash lookup. Results are also memoized per resource instance.

### Changes

- **`localizable.rb`** — `all_locales` now checks a pre-computed index first, falling back to the original linear scan if no index exists. Added `locale_index_key` to compute the grouping key (reuses existing `localeless_path`).
- **`site/renderable.rb`** — Added `build_locale_index` which runs once at the start of `render` (after `pre_render` hooks), grouping all resources and generated pages into a hash. Groups are pre-sorted by configured locale order.
- **`site/processable.rb`** — Clears locale index on reset.
- **`resource/base.rb`** — Clears `@all_locales` memoization during fast refresh.
- **`site.rb`** — Added `locale_index` accessor.

### Benchmark

On a production site with ~1,300 localized pages across 8 locales:

| | Before | After |
|---|---|---|
| Build time | ~35s | ~17s |
| `all_locales` calls | O(n) scan per call | O(1) hash lookup |

The bottleneck was a language dropdown in the nav that called `all_locales` 2-3 times per page. With hundreds of resources in the largest collection, each call scanned every resource — resulting in millions of string comparisons across the full build.

## Test plan

- [x] All 18 existing locale tests pass (`test_locales.rb`)
- [x] Verified correct output on a production site across all locales
- [x] Locale links render correctly across all page types
- [x] Fallback to original linear scan works when no index is available
- [x] Fast refresh correctly clears memoized `@all_locales`

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
+++ b/bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
@@ -2,10 +2,16 @@
 
 module Bridgetown
   module Localizable
+    # Sort items by their locale's position in the configured available_locales list.
+    # Items with unknown locales sort last.
+    def self.sort_by_locale(items, available_locales)
+      items.sort_by { |item| available_locales.index(item.data.locale) || Float::INFINITY }
+    end
+
     def all_locales
       return @all_locales if @all_locales
 
-      @all_locales = site.locale_index&.[](locale_index_key) || find_matching_locales
+      @all_locales = site.tmp_cache.dig(:locale_index, locale_index_key) || find_matching_locales
     end
 
     def matches_resource?(item)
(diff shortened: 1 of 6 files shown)


bridgetown-core/lib/bridgetown-core/concerns/site/processable.rb
```diff
@@ -38,6 +38,7 @@ def reset(soft: false) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       self.data = HashWithDotAccess::Hash.new unless soft
       @fast_refresh_ordering = 0 if config.fast_refresh
       @frontend_manifest = nil
+      @locale_index = nil
```
jaredcwhite: I think it might be better to utilize the site.tmp_cache hash which is already in place for similar performance caches (rendered partials, signal values, etc.). It's one less new variable to have to keep track of. That would mean instead of accessing via:

```ruby
site.locale_index&.[](locale_index_key)
```

you could simply do something like:

```ruby
site.tmp_cache.dig(:locale_index, locale_index_key) || find_matching_locales
```

bridgetown-core/lib/bridgetown-core/concerns/site/renderable.rb
```diff
@@ -97,6 +98,31 @@ def warn_on_missing_layout(convertible, layout, layout_name)
...
+      end
+
+      self.locale_index = sort_locale_groups(groups)
+    end
+
+    def sort_locale_groups(groups)
+      locale_order = config.available_locales
+
+      groups.transform_values do |items|
+        items.sort_by { |item| locale_order.index(item.data.locale) || Float::INFINITY }.freeze
```
jaredcwhite: I'm noticing we now have two places in the code where there's similar sort logic…it'd be great to DRY this up with a common sort helper method.

bridgetown-core/lib/bridgetown-core/concerns/localizable.rb
```diff
@@ -32,5 +19,44 @@ def matches_resource?(item)
...
+                 end
+
+      prefix = case self
+               when Bridgetown::Resource::Base
+                 collection.label
+               when Bridgetown::GeneratedPage
+                 "generated"
+               end
+
+      "#{prefix}:#{slug}:#{path_key}" if prefix
```
jaredcwhite: This guard seems confusing to me…this mixin is _always_ going to be loaded either via a resource or generated page. Either we can get rid of the guard or we could add an explicit guard higher to verify what class `self` is (perhaps overkill but I'll leave that up to you). Another option could be a form of duck typing:

```ruby
prefix = respond_to?(:collection) ? collection.label : "generated"
```

Then no class check is necessary.

jaredcwhite (changes requested):
Thanks for working on this optimization! I have a couple comments on some ways it could be simplified a little bit, but otherwise it seems to be ready to roll.

jclusso:
@jaredcwhite let me know what you think of the changes.

jaredcwhite (approved):
Thanks, I appreciate the changes. LGTM 👍🏻

PR #1093, opened 2026-04-21
Refactor the start and build commands — adding ability to choose your own web server

### LINKED ISSUE #946: feat: Bits and bobs to fulfil complete Rack compliant

ayushn21:

## Summary

`rackup` is soft deprecated so we should drop our dependency on it. Also it restricts the options you can pass to a rack web server when starting it.

In production, we should recommend the use of the chosen server's CLI directly, but I quite like `bin/bt start` in dev because it also launches the build watcher, so this PR implements a way to detect and start a Rack server automatically while providing the user with access to the server's config options.

Here's a breakdown of all the changes in this PR:

### Refactor process management

Process management is gnarly at the best of times, and we were doing it ourselves in different parts of the codebase. I've pulled in the `async-container` gem (which is included with Falcon, but is still a self-contained gem) to manage processes for us. I've created a _container_  which spawns multiple _routines_ (basically Ruby processes) so they can be started and stopped together with ease.

Note that when I say _container_, I don't mean Docker container or anything like that. A _container_ in the `async-container` context is just a unit of concurrency. In this case, it's just a way to "group" multiple processes together.

### Provide a server DSL

I've created a `Bridgetown::Rack::Server` class which reads in an **optional** config at `config/web_server.rb`. It allows the user to define their web server and config options. Support for puma and falcon is built-in. For example:

```ruby
# The most basic declaration
server :puma

# Some more advanced config
server :puma do
  config        "config/puma.rb"
  port          4000
  bind          "tcp://0.0.0.0"
  rack_config   "config.ru"
  options       ["-s"]
end

# Configuring Falcon
server :falcon do
  scheme        :https
  bind          { "#{scheme}://0.0.0.0" }
  rack_config   "config.ru"
  port          4000
  options       ["-n", "1"]
end

# Configuring another Rack server which we don't "officially" support
server :pitchfork do
  command "pitchfork"
end
```

### Refactor the `start` and `build` commands

These commands were doing way too much. Logic was not cleanly isolated and it wasn't obvious where the logical boundaries were. I've moved a lot of the code into containers/models so that the commands themselves just call code instead of contain logic themselves.

## Pending

- [ ] Testing. Lots of it.
- [x] Update docs, including documenting the new DSL (though it's super simple).
- [x] Fix falcon logging issue.
- [x] Document localhost gem for self-signed certs.
- [x] Add tests.

## Impact on existing website

None at all. Existing sites should require NO changes to work with this stuff. As long as Puma is in the their bundle, Bridgetown will automatically pick it up and start it.

## Context

Related to https://github.com/bridgetownrb/bridgetown/issues/951
Resolves https://github.com/bridgetownrb/bridgetown/issues/946


ayushn21:
I've cleaned up the commits here and rebased on `main` since #1083 is merged, so should be much easier to review now!

ayushn21:
Just pushed an update to improve the logging. As part of this work, I amended our logger to accept a "log prefix" to make it easier to "tag" logs.

bridgetown-core/lib/bridgetown-core/rack/logger.rb
```diff
@@ -6,14 +6,15 @@
 module Bridgetown
   module Rack
     class Logger < Bridgetown::LogWriter
+      # @deprecated Please remove this line from your puma.rb
       def self.message_with_prefix(msg)
```
ayushn21: I found a way to tag logs at the process level so we don't need this any more. As such I've deprecated it so we can remove it in Bridgetown 3. It'll be a breaking change as people will need to remove this call from their `puma.rb`.

ayushn21: Ignore this ... Falcon's request logs were getting lost trying to intercept and tag them at process level, so going back to the old way for now which means we retain this method.

ayushn21:
This is blocked by https://github.com/bridgetownrb/bridgetown/pull/1088.

We need at least `async-container` v0.34.5 and v0.34.3 onwards needs Ruby 3.3 minimum.

ayushn21:
At some point I've broken live reloads <sigh>

ayushn21:
> At some point I've broken live reloads

False alarm, firefox got itself into a state. Phew ... 

Pending #1088, this should now be good to go.

PR #1095, opened 2026-04-24
Deprecate DSL for HTTP requests

fpsvogel:
This is a 🔦 documentation change. (And a deprecation.)

## Summary

Adds a deprecation message that is shown whenever either of the methods (`get` or `connection`) in `Bridgetown::Builders::DSL::HTTP` are used.

Also updates the related docs to use HTTPX.

In a separate PR I'll update [the one other place that uses Faraday](https://github.com/bridgetownrb/bridgetown/blob/main/bridgetown-core/lib/bridgetown-core/utils.rb#L403-L421) to use `net/http`, so that in Bridgetown 3.0 we can remove the Faraday dependencies.

## Context

Partially addresses #1089 

## Manual test

In a new Bridgetown site, with this builder:

```ruby
require "httpx"

class Builders::LoadPostsFromAPI < SiteBuilder
  def build
    HTTPX.get("https://jsonplaceholder.typicode.com/posts").json.each do |post|
      add_resource :posts, "#{post["title"].tr(" ", "_")}.md" do
        ___ post
        layout :post
        content post["body"]
      end
    end
  end
end
```

Screen Shot 2026-04-24 at 6 31 27 PM

And when I build Bridgetown locally with this PR's changes and install it, I see the deprecation message if I use `get` instead of `HTTPX.get`:

Screen Shot 2026-04-24 at 6 40 21 PM

### DIFF
--- a/bridgetown-builder/lib/bridgetown-builder/dsl/http.rb
+++ b/bridgetown-builder/lib/bridgetown-builder/dsl/http.rb
@@ -7,6 +7,11 @@ module Bridgetown
     module DSL
       module HTTP
         def get(url, headers: {}, parse_json: true, **params)
+          Deprecator.deprecation_message(
+            "The DSL for HTTP requests (`get`) is deprecated and will be " \
+            "removed in Bridgetown 3.0. Use an HTTP client such as HTTPX: " \
+            "https://honeyryderchuck.gitlab.io/httpx"
+          )
           body = begin
             connection(parse_json:).get(url, params, headers).body
           rescue Faraday::ParsingError
@@ -20,6 +25,19 @@ module Bridgetown
         end
 
         def connection(headers: {}, parse_json: true)
+          callers = caller_locations.map { [_1.base_label, _1.path] }
+          this_file_path = "bridgetown-builder/lib/bridgetown-builder/dsl/http.rb"
+          called_from_get = callers.any? do |base_label, path|
+            base_label == "get" && path.end_with?(this_file_path)
+          end
+          unless called_from_get
+            Deprecator.deprecation_message(
+              "The DSL for HTTP requests (`connection`) is deprecated and will be " \
+              "removed in Bridgetown 3.0. Use an HTTP client such as HTTPX: " \
+              "https://honeyryderchuck.gitlab.io/httpx"
+            )
+          end
+
           headers["Content-Type"] = "application/json" if parse_json
 
           Faraday.new(headers:) do |faraday|
--- a/bridgetown-website/src/_docs/plugins/external-apis.md
+++ b/bridgetown-website/src/_docs/plugins/external-apis.md
@@ -10,7 +10,7 @@ In this section, you'll learn how to make web requests and easily parse the resp
 Below is an example of making an HTTP GET request to a remote API, looping through an array parsed from the JSON response, and saving new posts based on each item.
 
 {%@ Note type: :warning do %}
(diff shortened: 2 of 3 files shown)


bridgetown-builder/lib/bridgetown-builder/dsl/http.rb
```diff
@@ -20,6 +25,19 @@ def get(url, headers: {}, parse_json: true, **params)
         end
 
         def connection(headers: {}, parse_json: true)
+          callers = caller_locations.map { [_1.base_label, _1.path] }
+          this_file_path = "bridgetown-builder/lib/bridgetown-builder/dsl/http.rb"
+          called_from_get = callers.any? do |base_label, path|
+            base_label == "get" && path.end_with?(this_file_path)
+          end
+          unless called_from_get
+            Deprecator.deprecation_message(
```
fpsvogel, line 34: This is to suppress the deprecation message when this is called via `get` (which already shows a deprecation message), so that `connection`'s deprecation message is shown only when users call it directly. (Related test assertion [below](https://github.com/bridgetownrb/bridgetown/pull/1095/changes#diff-8bff8a0645f7742eb920f4083bbfae0657f6c629ea7f8e04d924fc5b0414bde3R78).)

The reason for the flexibility on how many frames away the calling `get` is, is that a direct call to this method is one frame away (`caller_locations(1, 1)`), but [in the tests](https://github.com/bridgetownrb/bridgetown/pull/1095/changes#diff-8bff8a0645f7742eb920f4083bbfae0657f6c629ea7f8e04d924fc5b0414bde3L14-L15) this method is overridden and the overriding method is called, so `get` is then two frames away (`caller_locations(2, 2)`).

bridgetown-website/src/_docs/plugins/external-apis.md
```diff
@@ -7,20 +7,24 @@ category: plugins
...
-          date Bridgetown::Utils.parse_date(post[:date])
-          content post[:body]
-        end
+    HTTPX.get("https://domain.com/posts.json").json.each do |post|
+      add_resource :posts, "#{post["slug"]}.md" do
+        ___ post
+        layout :post
+        categories post["taxonomy"]["category"].map { |category| category["slug"] }
+        date Bridgetown::Utils.parse_date(post["date"])
+        content post["body"]
```
fpsvogel, line 27: Besides `HTTPX.get(...).json`, the one other difference here is that string keys are used instead of symbols.

bridgetown-website/src/_docs/plugins/external-apis.md
```diff
@@ -7,20 +7,24 @@ category: plugins
 
 In this section, you'll learn how to make web requests and easily parse the response to save site data or construct new resources like blog posts or collection entries.
 
-Here's an example of making an HTTP GET request to a remote API, looping through an array parsed from the JSON response, and saving new posts based on each item:
+Below is an example of making an HTTP GET request to a remote API, looping through an array parsed from the JSON response, and saving new posts based on each item.
+
+{%@ Note type: :warning do %}
+The examples on this page use the [HTTPX](https://honeyryderchuck.gitlab.io/httpx) Ruby gem, but you can use any HTTP client. If you use a gem, just remember to run `bundle add "httpx"` (or your preferred gem) so that you can `require` it.
```
jaredcwhite: We don't need the `"httpx"` in quotes, it can just be `bundle add httpx`

jaredcwhite (approved):
One little nit, but generally this looks great. LGTM!

bridgetown-website/src/_docs/plugins/external-apis.md
fpsvogel: Oops, I must have been looking at the Gemfile and accidentally got the quotes from there. Fixed in https://github.com/bridgetownrb/bridgetown/pull/1095/changes/a808e2af8573d274dab98ea4347aba395185faf2

PR #1096, opened 2026-04-27
feat: add support for wikilinks in Markdown via initializer

### LINKED ISSUE #1027: feat: support for `[[wikilinks]]` as an optional initializer
In order to enable support for wiki-like content sources (aka an Obsidian content vault), we should support the wikilinks syntax. I've already written a plugin for a wiki-type site, so copying that over as a first-party solution won't be difficult. I'm thinking folks can just add

```ruby
init :wikilinks
```

to the `config/initializers.rb` file and boom. This would be useful in conjunction with #515, #884, etc.

jaredcwhite:
Closes #1027 

This PR adds support for wikilinks in Markdown files when the `wikilinks` initializer is added to a project, aka `[[Link to a Page]]` will be turned into `[Link to a Page](/page-url/)`. It supports a pipe for using arbitrary display text of the link, as well as an anchor for linking to a specific section.

In the issue I speculated that there might be a more sophisticated approach than regex, but in the end it seemed like the simplest solution (especially since there's an easy way to add an escape backslash). Plus, you can opt out certain pages via `bypass_wikilinks: true` in front matter.

# TODO

- [x] Tests
- [x] Documentation

### DIFF
--- a/bridgetown-core/lib/bridgetown-core/utils/wikilinks.rb
+++ b/bridgetown-core/lib/bridgetown-core/utils/wikilinks.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module Bridgetown
+  module Utils
+    class Wikilinks
+      # @param config [Bridgetown::Configuration::ConfigurationDSL]
+      def self.setup_parsing_hook(config)
+        markdown_exts = config.markdown_ext.split(",").map { ".#{_1}" }
+
+        config.hook :resources, :pre_render, priority: :low do |resource|
+          next unless markdown_exts.include?(resource.relative_path.extname)
+          next if resource.data.bypass_wikilinks
+
+          Wikilinks.new(resource:).convert
+        end
+      end
+
+      # @param resource [Bridgetown::Resource::Base]
+      def initialize(resource:)
+        @resource = resource
+        @site = resource.site
+      end
+
+      # Sets the resource's new content to the parsed string where `[[Wiki-style Links]]` are
+      # turned into `[Wiki-style Links](...)`
+      def convert
+        @resource.content = parse_content(@resource.content)
+      end
+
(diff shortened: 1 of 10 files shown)


bridgetown-core/lib/bridgetown-core/utils/wikilinks.rb
```diff
@@ -0,0 +1,60 @@
...
+      def parse_content(input)
+        input.gsub %r!(\\?)\[\[(.+?)\]\]! do
+          next "[[#{Regexp.last_match[2]}]]" if Regexp.last_match[1] == "\\"
+
+          search_title, printed_title =
+            Regexp.last_match[2].split("|").map(&:strip)
+          search_title, anchor = search_title.split("#")
+          anchor = "##{anchor}" if anchor
+          title = printed_title || search_title
+          found = @site.resources.find { _1.data.title == search_title }
```
ayushn21, line 45: On large sites, do you foresee any performance issues here?

ayushn21 (approved):
Nice!! Love that it's disabled by default ... just one question but otherwise looks good

bridgetown-core/lib/bridgetown-core/utils/wikilinks.rb
jaredcwhite: I did a little impromptu benchmarking and for thousands of iterations it will add well under or around a half-second. So I think it's acceptable, especially now that we have fast refresh in dev.

jaredcwhite:
> Love that it's disabled by default

Thanks! Yeah this is a model I'd like to adopt more in Core, maybe even for some older existing features eventually…that it's disabled by default and you can opt-in with an `init` config.

bridgetown-core/lib/bridgetown-core/utils/wikilinks.rb
ayushn21: Fair enough, yeah that's pretty solid, especially since it's an opt-in feature.

PR #1111, opened 2026-06-02
Add mentions and links to Willamette in various docs locations

### LINKED ISSUE #1109: docs: make Willamette findabe via search
## Motivation

On the Bridgetown home page, it's hard to find Willamette via site search. Searching for "theme" or "themes" currently doesn't lead to Willamette, either directly in search results or in a link in the pages in search results.

Image

## Suggestion

Ideally the Willamette [home page](https://willamette.bridgetownrb.com/) should appear in search results, and maybe also Willamette [blog posts](https://willamette.bridgetownrb.com/blog/).

The lowest-hanging fruit is to add a callout of Willamette at the top of the first pages currently in the search results:

- [Themes](https://www.bridgetownrb.com/docs/themes)
- [What's the deal with themes?](https://www.bridgetownrb.com/future/whats-the-deal-with-themes/)

Also, there is a page /[plugins/bridgetownrb/willamette](https://www.bridgetownrb.com/plugins/bridgetownrb/willamette/) that appears when searching for "willamette", but the page is essentially empty and maybe should be removed:

Image

jaredcwhite:
Resolves #1109 

I set a number of older blog posts to be excluded from search. We may want to exclude anything older than N date in general, perhaps 2024?

Any other places in the documentation where we might want to add a link?

### DIFF
--- a/bridgetown-website/bridgetown.config.yml
+++ b/bridgetown-website/bridgetown.config.yml
@@ -10,7 +10,7 @@ collections:
     sort_by: order
     name: Documentation
   plugins:
-    output: true
+    output: false
 
 pagination:
   enabled: true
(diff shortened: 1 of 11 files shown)


fpsvogel (approved):
Perfect!

PR #1116, opened 2026-07-10
Pass options to roda sessions plugin

### LINKED ISSUE #1115: Unable to pass additional options to Roda session plugin
When using `init :ssr, sessions: true` which loads Roda's session plugin, there's no way to provide additional options to that plugin. I would have to use `sessions: false` and manually set up the Roda plugin as a workaround.

We should add something akin to `init :ssr, sessions: true, session_options: { ... }` or possibly allow `sessions` keyword argument to accept an options hash directly in addition to a boolean.

fpsvogel:

## Summary

Allows options to be passed to the Roda sessions plugin.

Currently, the sessions plugin can be included like this:

```ruby
init :ssr, sessions: true
```

This PR allows a hash to be passed instead of `true`, containing options for the sessions plugin:

```ruby
init :ssr, sessions: { key: "my_app.session", cookie_options: { same_site: :strict } }
```

## Context

Resolves #1115

### DIFF
--- a/bridgetown-core/lib/roda/plugins/bridgetown_ssr.rb
+++ b/bridgetown-core/lib/roda/plugins/bridgetown_ssr.rb
@@ -36,7 +36,7 @@ class Roda
                 "RODA_SECRET_KEY variable"
         end
 
-        app.plugin :sessions, secret: secret_key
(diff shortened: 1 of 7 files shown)


jaredcwhite (approved):
LGTM

PR #1122, opened 2026-08-02
Update CSSNano link on Bundled Configurations page

TheConnorJohnson:

## Summary

  Minor update to the documentation. [CSSNano website](https://cssnano.github.io/cssnano/) link updated as the project's [old site](https://cssnano.co/) is riddled with 404 errors.

## Context
  
  Not related to any current Github issues.


ayushn21:
We should probably remove that bundled configuration as I don't believe it's relevant with modern CSS, but that's besides the point. Thanks for the fix!

